PR #59, opened 2021-03-19
INTYGFV-13827: Added tests for input components

mhornfeldt:


mww-sopra-steria (changes requested):
Good tests! Thinking if we should add two more type of tests that are important for the behavior of the components.
1) That the component gets the correct label.
2) That it is possible to select/deselect the component through its label. Now it "only" use getByRole/queryAllByRole.

mww-sopra-steria (approved):

PR #65, opened 2021-04-16
INTYGFV-13570: Added renew function for lisjp

mhornfeldt:


packages/common/src/types/certificate.ts
```diff
@@ -358,6 +359,8 @@ export enum CertificateEventType {
   REVOKED = 'REVOKED',
   REPLACED = 'REPLACED',
   REPLACES = 'REPLACES',
+  RENEWS = 'RENEWS',
```
mww-sopra-steria, line 362: What do you think about RENEWAL_OF instead of RENEWS?

mww-sopra-steria (approved):
Approved! Consider comment about naming.

mww-sopra-steria (approved):
Just have some concerns surrounding the body-field in ResourceLink. Continue with this for now. When we have updated the Webcert-backend we will take a look at possible options.

PR #66, opened 2021-04-22
INTYGFV-13569: Added functionality for sending certificates

mhornfeldt:


mww-sopra-steria (approved):
For now it is OK that the user text is hardcoded in ShowHistory. 

But as with many other certificate type specific situations we need to provide the client with text resources from the backend to handle this better.

PR #82, opened 2021-06-07
INTYGFV-13633: Added header for question and changed size of title if…

mhornfeldt:
… there is a header

### DIFF
--- a/packages/common/src/components/utils/Accordion.tsx
+++ b/packages/common/src/components/utils/Accordion.tsx
@@ -48,17 +48,9 @@ const Accordion: React.FC<Props> = ({ title, description, additionalStyles, disp
 
   const getHeader = () => {
     if (!hasHeader) {
-      return (
-        <>
-          <h4 className={`iu-fs-300 iu-color-black ${additionalStyles}`}>{getContents()}</h4>
-        </>
-      )
+      return <h4 className={`iu-fs-300 iu-color-black ${additionalStyles}`}>{getContents()}</h4>
     } else {
-      return (
-        <>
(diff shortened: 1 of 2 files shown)


packages/common/src/components/utils/Accordion.tsx
```diff
@@ -36,9 +38,33 @@ const Accordion: React.FC<Props> = ({ title, description, additionalStyles, disp
     item.classList.toggle('ic-expandable--expanded')
   }
 
-  return (
-    <div className="ic-expandable" ref={expandableRef}>
-      <h4 className={`iu-fs-300 iu-color-black ${additionalStyles}`}>
+  const Text = styled.p`
```
RobertDanielsson, line 41: Perhaps this styled component should be declared outside of the component?

packages/common/src/components/utils/Accordion.tsx
```diff
@@ -36,9 +46,25 @@ const Accordion: React.FC<Props> = ({ title, description, additionalStyles, disp
     item.classList.toggle('ic-expandable--expanded')
   }
 
-  return (
-    <div className="ic-expandable" ref={expandableRef}>
-      <h4 className={`iu-fs-300 iu-color-black ${additionalStyles}`}>
+  const getHeader = () => {
+    if (!hasHeader) {
+      return (
+        <>
```
RobertDanielsson, line 52: Are these fragments needed here? (The empty <>). Just asking since there's just one line between them (same on line 59). Perhaps I'm missing something though

RobertDanielsson (approved):

PR #89, opened 2021-06-21
INTYGFV-13750: Updated send certificate functions to go through the f…

mhornfeldt:
…acade

### DIFF
--- a/packages/common/src/types/resourceLink.ts
+++ b/packages/common/src/types/resourceLink.ts
@@ -6,10 +6,6 @@ export interface ResourceLink {
   enabled: boolean
 }
 
-export interface ResourceLinkSend extends ResourceLink {
-  modalBody: string
-}
-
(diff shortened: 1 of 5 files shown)


mww-sopra-steria (changes requested):
I guess there have to be some kind of change to handle receiver. Maybe for now just remove the receiver from the confirmButtonText as it is redundant or see first what solution that comes from the backend...

packages/common/src/types/resourceLink.ts
```diff
@@ -8,7 +8,6 @@ export interface ResourceLink {
 
 export interface ResourceLinkSend extends ResourceLink {
```
mww-sopra-steria, line 9: Could we remove this specific ResourceLink as the body is already part of the generic ResourceLink?

mww-sopra-steria (changes requested):

mww-sopra-steria (approved):

PR #96, opened 2021-06-24
INTYGFV-13633: Updated text alignment

mhornfeldt:


mww-sopra-steria (approved):
Could be a little bit more specific in your commit comment. Text alignment "for what".

PR #105, opened 2021-07-01
INTYGFV-13550: Validation message should not be visible for every che…

mhornfeldt:
…ckbox and only correct values should be saved


mww-sopra-steria (changes requested):
Can you see if you can introduce some unit tests to validate the new logic that you have added to the component?

mww-sopra-steria (approved):
Looks very good!

PR #124, opened 2021-09-14
INTYGFV-13633: Updated styling, fixed new header and added more texts for events

mhornfeldt:


packages/webcert/src/feature/certificate/CertificateHeader/ShowHistory.tsx
```diff
@@ -155,15 +168,21 @@ const ShowHistory: React.FC<Props> = ({ historyEntries, certificateMetadata }) =
     }
```
mww-sopra-steria, line 168: Same as the comment on Webcert. Don't we need to add "complemented"?

mww-sopra-steria (changes requested):
Have a comment about "complemented".

Also have a look and see if you can add some tests to validate that the new header presents the correct information. It is good to have for later also when we gonna extend with more information/functionality in the header.

mww-sopra-steria (approved):

PR #136, opened 2021-10-18
INTYGFV-14015: Cypress tests added working pipeline

morongon:


mww-sopra-steria (approved):
We might do some adjustments when it comes to executing the tests in the pipeline. But I think it is best to do them in a separate branch (and PR).

PR #139, opened 2021-10-20
INTYGFV-14152: Refactored certificate side panel, removed minimize feature, styling adjustments 

RobertDanielsson:
Removed unused props and code. The tabs and tab content is now grouped together. Earlier they were dependent on having the same if case order, otherwise there would be a mismatch between tabs and their content


mww-sopra-steria (commented):
Is it possible to add some tests to the side panel? It has some logic to which tabs to add for example.

PR #157, opened 2021-11-01
INTYGFV-14196: Added highlight validation

mhornfeldt:


mww-sopra-steria (changes requested):
Suggest that you add tests for the new logic in validationUtils. 

I know we are lacking tests for the certificateMiddleware, but see if you can add tests that make sure that the style is toggled correctly if the highlight-rule exists.

packages/common/src/utils/validationUtils.test.ts
```diff
@@ -1027,4 +1027,44 @@ describe('Set initial values to a certificate', () => {
       expect(certificate.data['1.2'].visible).toBe(true)
     })
   })
+
+  describe('Validate highlight rule for boolean value', () => {
```
mww-sopra-steria, line 1031: These tests doesn't test the added logic. Instead it tests the parseExpression, which is already verified by other tests that it can evaluate a boolean expression.

packages/webcert/src/store/certificate/certificateMiddleware.test.ts
```diff
@@ -147,6 +149,28 @@ describe('Test certificate middleware', () => {
...
+    })
+
+    it('shall unstyle certificate data element', async () => {
+      const certificate = getCertificateWithHiglightValidation('NOT_VALUE')
+
+      testStore.dispatch(updateCertificate(certificate))
+      testStore.dispatch(validateCertificateInFrontEnd(certificate.data[0]))
+
+      await flushPromises()
+      expect(testStore.getState().ui.uiCertificate.certificate.data[0].style).toEqual(CertificateDataElementStyleEnum.HIGHLIGHTED)
```
mww-sopra-steria, line 170: Should be asserted as NORMAL.

mww-sopra-steria (changes requested):

mww-sopra-steria (approved):
Looks good!

PR #162, opened 2021-11-03
INTYGFV-13878: Fixed bug where the wrong complement question was dispatched as handled

RobertDanielsson:


mww-sopra-steria (approved):
Nice!

PR #164, opened 2021-11-03
INTYGFV-14186: Added character counter

mhornfeldt:


mww-sopra-steria (changes requested):
The new limit component should have unit tests for the logic it contains. 

Also see if any other components changed within the PR has added logic that could be validated by a init test.

mhornfeldt:
This PR should not be reviewed implementation is not done.

mww-sopra-steria (approved):

PR #165, opened 2021-11-03
INTYGFV-13877: Created complement status and adjusted sent status to also check if complements exists

RobertDanielsson:


mww-sopra-steria (changes requested):
Nice! 

Suggests that you add unit test for the new logic in the certificateUtil, even if it is tested indirectly through the components. You can include validation of negative scenarios.

mww-sopra-steria (approved):

PR #178, opened 2021-11-11
INTYGFV-14207: Added protected person for doctor and about wc modal

mhornfeldt:


packages/webcert/src/store/utils/utilsMiddleware.ts
```diff
@@ -13,11 +13,22 @@ import {
   updateDiagnosisTypeahead,
   updateDynamicLinks,
 } from './utilsActions'
+import { updateCertificate } from '../certificate/certificateActions'
+
+const handleUpdateCertificate: Middleware<Dispatch> = ({ dispatch, getState }) => (next) => (action: AnyAction): void => {
+  next(action)
```
mww-sopra-steria, line 19: Suggest that you remove handleUpdateCertificate and instead dispatch getAllDynamicLinks when the app loads in App.tsx.

mww-sopra-steria (changes requested):

mww-sopra-steria (approved):

PR #188, opened 2021-11-15
INTYFV-14134: Added errorId text to error modal and error page with ability to copy the error id.

RobertDanielsson:
* I will refactor this after tomorrows demo, copy functionality is in place but i'd like to add a text that shows that the text is copied. Wanted something in place for the demo to display the errorId


mww-sopra-steria (approved):
Great!

PR #199, opened 2021-11-18
INTYGFV-14183: Fixed dynamic height of certificate so content doesn't disappear if header expands

RobertDanielsson:


mww-sopra-steria (approved):
Nice!

PR #220, opened 2021-11-23
INTYGFV-14257: Updated modal text when certificate is available for p…

mhornfeldt:
…atient


mww-sopra-steria (approved):
This is something to look at after the pilot. On how we get rid of certificate types hardcoded texts like this. But I think the solution you have implemented right now is acceptable.

PR #238, opened 2021-12-07
INTYGFV-14134: Created functionBlockers in redux slices to block functionality while waiting for api calls to be handled

RobertDanielsson:
I've added blocking of buttons to ICF, Question and Certificate but not FMB (Not sure if needed somewhere). I probably missed some button, but it's quick and easy to add to more. Just use the correct selector, for example isCertificateFunctionDisabled for certificate functionality blocking


RobertDanielsson:
Need to fix failing tests.. :-)

packages/common/src/utils/certificateUtils.ts
```diff
@@ -111,3 +111,7 @@ export const getComplementedByCertificateRelation = (certificateMetadata: Certif
     (r) => r.type === CertificateRelationType.COMPLEMENTED && r.status !== CertificateStatus.REVOKED
   )
 }
+
+export interface FunctionDisabled {
```
mww-sopra-steria, line 115: I think we should move this to another file. Either something already existing or create a new file for this time of functionality.

packages/webcert/src/components/question/QuestionForm.tsx
```diff
@@ -41,10 +41,10 @@ const QuestionForm: React.FC<Props> = ({ questionDraft }) => {
   const isSaved = useSelector(isQuestionDraftSaved)
   const isMissingType = useSelector(isQuestionMissingType)
   const isMissingMessage = useSelector(isQuestionMissingMessage)
-  const isSending = useSelector(isSendingQuestion)
   const showValidationMessages = useSelector(isDisplayValidationMessages)
   const [message, setMessage] = useState(questionDraft.message)
   const subjects: QuestionType[] = Object.values(QuestionType)
+  const isFunctionBlocked = useSelector(isQuestionFunctionDisabled)
```
mww-sopra-steria, line 47: For consistency it should be called isFunctionDisabled instead of isFunctionBlocked.

packages/webcert/src/components/utils/functionBlockerUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+export const isFunctionBlocked = (functionBlockers: FunctionBlocker[]): boolean => {
+  return functionBlockers.length > 0
+}
+
+export const generateFunctionBlocker = (): FunctionBlocker => {
+  return { id: uuidv4() }
+}
+
+// https://stackoverflow.com/questions/105034/how-to-create-guid-uuid
+function uuidv4(): string {
```
mww-sopra-steria, line 38: Simplify to just use milliseconds for id. No need to be globally unique between clients.

packages/webcert/src/feature/certificate/CertificateHeader/CertificateHeader.tsx
```diff
@@ -16,7 +16,7 @@ import styled from 'styled-components/macro'
 import { Divider } from '@frontend/common'
 import CreateCertificateFromCandidateModal from '../Modals/CreateCertificateFromCandidateModal'
 import { resourceLinksAreEqual, ResourceLinkType } from '@frontend/common/src'
-import { getQuestions } from '../../../store/question/questionSelectors'
+import { getQuestions, isQuestionFunctionDisabled } from '../../../store/question/questionSelectors'
```
mww-sopra-steria, line 19: Should this really be isQuestionFunctionDisabled? The CertificateHeader should enable/disable based on the certificate.

packages/webcert/src/store/api/apiMiddleware.test.ts
```diff
@@ -150,4 +150,23 @@ describe('Test API middleware', () => {
       expect(throwErrorAction?.payload.type).toEqual(ErrorType.SILENT)
     })
   })
+
+  describe('handle functionBlocker', () => {
```
mww-sopra-steria, line 154: Would be good to include tests that verifies that the functionalBlockerType is called as "a pair" no matter if the api call is a success or it causes an error.

The implementation in the apiMiddleware looks solid, but it is good if the tests verifies that no one breaks this important logic.

packages/webcert/src/store/api/apiMiddleware.ts
```diff
@@ -4,19 +4,25 @@ import { apiCallBegan, apiCallFailed, apiCallSuccess, ApiError, apiGenericError,
...
+import { FunctionBlocker, generateFunctionBlocker } from '../../components/utils/functionBlockerUtils'
 
 const handleApiCallBegan: Middleware = ({ dispatch }: MiddlewareAPI) => (next: Dispatch) => async (action: AnyAction) => {
   if (!apiCallBegan.match(action)) {
     return
   }
 
-  const { url, method, data, onStart, onSuccess, onError, onArgs } = action.payload
+  const { url, method, data, onStart, onSuccess, onError, onArgs, functionBlockerType } = action.payload
+  const functionBlocker: FunctionBlocker = generateFunctionBlocker()
```
mww-sopra-steria, line 15: Only generate a functionBloocker if it is necessary. Now it will generate one even if there is no functionBlockerType.

mww-sopra-steria (changes requested):
Nice! A few comments to look at, but mostly minor things.

I think we need to let Test try it out a little to see if we have covered all actions that should be toggled, but I believe you have covered the majority if not all.

packages/webcert/src/components/utils/functionBlockerUtils.ts
RobertDanielsson: I tried changing to ms but my tests started failing, is my compiler too fast? 🤔

RobertDanielsson: (When i created two function blockers they got the same id)

mww-sopra-steria (approved):

PR #239, opened 2021-12-08
INTYGFV-14277: Should validate disable when certificate is loaded

mhornfeldt:


packages/common/src/utils/validationUtils.ts
```diff
@@ -236,6 +237,25 @@ function shouldBeDisabled(metadata: CertificateMetadata) {
   return metadata.status === CertificateStatus.LOCKED || metadata.status === CertificateStatus.LOCKED_REVOKED
 }
 
+export function setDisableForChildElement(data: CertificateData, validationResult: ValidationResult) {
```
mww-sopra-steria, line 240: I need to think about this a bit. Especially consider if we can make this more generic. Right now it is hardcoded for ConfiguUeCheckboxMultipleCodes, but the disable validation should generic to be used for other configs.

mww-sopra-steria (commented):
Will look at this during today. See comment.

PR #241, opened 2021-12-08
INTYGFV-14280: Label and header should be set in question component for read only view

mhornfeldt:


packages/webcert/src/feature/certificate/Question/Question.tsx
```diff
@@ -126,8 +132,9 @@ const Question: React.FC<QuestionProps> = ({ id }) => {
   )
 
   function getQuestionComponent(config: CertificateDataConfig, displayMandatory: boolean, readOnly: boolean) {
+    console.log(getHeading())
```
mww-sopra-steria, line 135: Remove console-log

mww-sopra-steria (approved):
Just remove console.log before merging.

PR #246, opened 2021-12-10
INTYGFV-14131: Created new UeRadioGroupOptionalDropdown which has the option to display a dropdown after the radio button. Added hidden style to hide questions

RobertDanielsson:
Plus some styling to place the dropdown correctly after the radio button


mww-sopra-steria (approved):
Nice! If you think you can add some tests, that would be great! It would be good to have this merged early after lunch so that Test can use the changes for there tests in the afternoon.

RobertDanielsson:
> Nice! If you think you can add some tests, that would be great! It would be good to have this merged early after lunch so that Test can use the changes for there tests in the afternoon.

I'll merge now and add tests afterwards if I can find any good tests to implement, just to have the new version live :-)

PR #250, opened 2021-12-14
INTYGFV-14284: Changed CopyCertificateButton isReplaced check to check if child has relatio…

RobertDanielsson:
…n type COPIED instead of REPLACED


mww-sopra-steria (changes requested):
Remove logic and tests that are related to the "previous" implementation.

mww-sopra-steria (approved):
Nice! Don't forget to change the commit message before committing :-)

PR #264, opened 2022-01-05
INTYGFV-14292: Added protected person agreement modal

mhornfeldt:


mww-sopra-steria (changes requested):
See the comment on the Webcert PR (https://github.com/sklintyg/webcert/pull/621).

mww-sopra-steria (dismissed):
Looks good! See comment about moving the logic that now sits inside the CertificateHeader.

karlssone (approved):

PR #267, opened 2022-01-11
INTYGFV-14347: Footer of side panel should not hide complements or qu…

mhornfeldt:
…estions


mww-sopra-steria (approved):
What is the 55 pixels based on? Will it stay 55px or is there a risk that it must be dynamic?

PR #268, opened 2022-01-11
INTYGFV-14356: FMB should not be updated if code system is not icd10

mhornfeldt:


mww-sopra-steria (changes requested):
I think some tests for the component and the middleware would be great :-)

karlssone (approved):

PR #277, opened 2022-01-20
INTYGFV-14330: Added possibility for components to dispatch client validation errors

mhornfeldt:


mhornfeldt:
PR is draft should not be reviewed

packages/common/src/components/Inputs/DatePickerCustom/DatePickerCustom.tsx
```diff
@@ -6,6 +6,8 @@ import { _dateReg, _format, formatDateToString, getValidDate, QuestionValidation
 import { DatePickerWrapper, StyledButton, TextInput, ValidationWrapper, Wrapper } from './Styles'
 import calendarImage from '../../../images/calendar.svg'
 import 'react-datepicker/dist/react-datepicker.css'
+import { updateClientValidationError } from '@frontend/webcert/src/store/certificate/certificateActions'
```
mww-sopra-steria, line 9: We have to avoid dependencies from common to webcert. See if you can move something more to common so we can get rid this dependency to webcert.

packages/webcert/src/store/certificate/certificateSelectors.ts
```diff
@@ -148,6 +148,29 @@ export const getCareUnitValidationErrors = () => (state: RootState): ValidationE
   return state.ui.uiCertificate.certificate.metadata.careUnitValidationErrors
 }
 
+const doesFieldsMatch = (payloadField: string, validationField: string) => {
+  return !validationField || validationField.includes(payloadField)
+}
+
+export const getVisibleValidationErrors = (questionId: string, field: string) => (state: RootState): ValidationError[] => {
```
mww-sopra-steria, line 155: For performance reason memoize the clientValidation errors so that all the filtering logic doesn't have to be done if there is no change.

Here is a link: https://react-redux.js.org/api/hooks#using-memoizing-selectors

mww-sopra-steria (changes requested):
Great job! A couple of comments that you should take a look at before merging.

packages/webcert/src/store/certificate/certificateSelectors.ts
mhornfeldt: Is it possible to do this when we need attributes from the component to do the filtering?

mhornfeldt: @mww-sopra-steria

mhornfeldt: @mww-sopra-steria Is it really possible to do use memoize when we need attributes from the component to do the filtering? If we don't have the questionId and field we can't filter out the validationMessages for the specific component

mww-sopra-steria: I turn that question around a little. Have you found any documentation or tried and found that it won't work?

mhornfeldt: @mww-sopra-steria https://stackoverflow.com/questions/40291084/use-reselect-selector-with-parameters

mww-sopra-steria: It seems like there are ways to implement it, but I don't find it so intuitive or easy. I suggest you merge the PR and then I will analyze what the possible impact the current implementation has. If it is none to minimal, we can be satisfied with the current selector. If not it will be easier to experiment with the client by having all the changes integrated in the master branch.

mww-sopra-steria (approved):
See my last comment relating to the selector and memoization.

PR #278, opened 2022-01-20
INTYGFV-14360 Disable buttons when sending answer requests

perfli:


packages/webcert/src/components/question/QuestionItem.test.tsx
```diff
@@ -244,6 +245,37 @@ describe('QuestionItem', () => {
...
+
+      userEvent.click(sendButton)
+
+      expect(sendButton).toBeDisabled()
+      expect(cancelButton).toBeDisabled()
+
+      await flushPromises()
+    })
+
+    it('disable send and cancel when deletet answer draft is being sent', async () => {
```
mhornfeldt, line 264: Deleted instead of deletet

PR #292, opened 2022-02-04
INTYGFV-14419: Certificate should be reloaded when handling complemen…

mhornfeldt:
…t since new resource links could be available


mww-sopra-steria (changes requested):
Missing unit test for the added behavior.

mww-sopra-steria (approved):

PR #294, opened 2022-02-07
INTYGFV-14447: All update functions of the diagnosis component should go through the same debounce

mhornfeldt:


mww-sopra-steria (approved):
Nice!

PR #306, opened 2022-02-14
INTYGFV-14412: Activate icon wrap should be set correctly

mhornfeldt:


packages/common/src/components/utils/InfoBox.tsx
```diff
@@ -61,7 +61,7 @@ const InfoBox: React.FC<Props> = ({ type, children, additionalStyles, squared, a
     // @ts-expect-error squared is giving error but it's working as intended
     <StyledWrapper squared={squared} className={`ic-alert ic-alert--status iu-lh-body ${getWrapperClass()} ${additionalStyles}`}>
       <Wrapper className={`${!activateIconWrap ? 'iu-flex' : ''} ${additionalWrapperStyles}`}>
-        <Icon activateIconWrap className={`${!activateIconWrap ? '' : 'iu-fl'} ic-alert__icon ${getIconClass()}`} />
+        <Icon activateIconWrap={!!activateIconWrap} className={`${!activateIconWrap ? '' : 'iu-fl'} ic-alert__icon ${getIconClass()}`} />
```
perfli, line 64: No issue really, but it feels a bit off to have `!!` on a boolean prop. I'm guessing you need to have it since `activateIconWrap` might be undefined if not set on component? My preference in that case would be to set a default value of `false` , i.e.
```typescript
const InfoBox: React.FC<Props> = ({ ..., activateIconWrap = false }) => {
```
and reverting
```javascript
<Icon activateIconWrap={!!activateIconWrap} />
```
to
```javascript
<Icon activateIconWrap />
```

perfli (approved):

packages/common/src/components/utils/InfoBox.tsx
mhornfeldt: I don't think only using activateIconWrap is correct here. Doesn't this translate to activateIconWrap={true}?

perfli: Sorry, my bad. I misremembered how passing props to components works in React. For my example to work you would have to do
```javascript
<Icon activateIconWrap={activateIconWrap} />
```

PR #312, opened 2022-02-17
INTYGFV-14481: Added error message for when question cannot be sent

mhornfeldt:


packages/webcert/src/components/error/modals/SendQuestionProblem.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import WCDynamicLink from '../../utils/WCDynamicLink'
+import { ModalProps } from './errorUtils'
+import ErrorModalBase from './ErrorModalBase'
+
+export const GENERAL_ERROR_TITLE = 'Meddelandet har inte skickats'
+
+const GeneralErrorReload: React.FC<ModalProps> = ({ errorData }) => {
```
perfli, line 8: The variable name doesn't make sense to me, is it copy-paste?

PR #318, opened 2022-03-02
INTYGFV-14203: Added possibility to show system banners

mhornfeldt:


packages/common/src/components/utils/InfoBox.tsx
```diff
@@ -28,9 +28,10 @@ interface Props {
   additionalWrapperStyles?: string
   squared?: boolean
   activateIconWrap?: boolean
+  isGlobal?: boolean
 }
 
-const InfoBox: React.FC<Props> = ({ type, children, additionalStyles, squared, additionalWrapperStyles, activateIconWrap }) => {
+const InfoBox: React.FC<Props> = ({ type, children, additionalStyles, squared, additionalWrapperStyles, activateIconWrap, isGlobal }) => {
```
perfli, line 34: I feel like we´re trying to force too much functionality into this component, making the solution more complex. From a SOLID perspective it's violating the SRP principle. Will a `SystemBanner` ever need all the functionality of an `InfoBox`? Will any other notification component need the global property of an `InfoBox`? If not then perhaps the design would be better off with a specialized component. With React it's fairly cheap to create components and I think it would be beneficial to prefer several specialized, slightly similar, components over fewer god components with lots of configuration

packages/webcert/src/feature/certificate/NotificationBanners/SystemBanner.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react'
+import { useSelector } from 'react-redux'
+import NotificationBannerBase from './NotificationBannerBase'
+import { getBanners } from '../../../store/utils/utilsSelectors'
+import { Banner } from '../../../store/utils/utilsReducer'
+
+const SystemBanner: React.FC = () => {
```
perfli, line 7: Just a minor thing, but technically this component renders multiple banners and the name `SystemBanners` would convey that better

perfli (changes requested):

packages/common/src/components/utils/InfoBox.tsx
mhornfeldt: Most of the props sent to the component refer to styling. This is also the case for the newly added global property. Since the property refers to styling and not functionality I don't agree that it violates SRP. The isGlobal property determines if the info box should have the UX-framework defined global styling. Since system banners are specific to the notifications added from the IA application there could be other banners which also have the global styling. By creating another InfoBox component for the global look we would create duplicated code.

perfli: The responsibility lies with determining which class to apply based on properties. Consider the alternative; a two single purpose components, e.g. `<InfoBox />` and `<GlobalInfoBox />`. In fact, since the properties refers to styling I would argue that the duplicated code will be minimal and trivial. It's already abstracted to CSS classes.

perfli: To clarify, splitting the component into `<InfoBox />` and `<GlobalInfoBox />` was just an example, it's not the solution I'm suggesting.

perfli (changes requested):
Should system banners also be displayed on the `ErrorPage`? If so, then perhaps the `SystemBanners` component should be moved from `feature/certificate` to a more general place.

packages/webcert/src/page/CertificatePage.tsx
```diff
@@ -72,12 +72,12 @@ const CertificatePage: React.FC = () => {
     <Root>
       <CertificateDeletedModal routedFromDeletedCertificate={routedFromDeletedCertificate} />
       <NoFlexGrow>
-        <SystemBanners />
         <AppHeader
           logo={logo}
           alt={'Logo Webcert'}
           primaryItems={[<WebcertHeaderUser />, <WebcertHeaderUnit />]}
           secondaryItems={secondaryItems}
+          banners={[<SystemBanners />]}
```
perfli, line 80: Why pass component as prop instead of including it in `AppHeader`? I think that would make more sense since all pages will have header and system banners components.

mhornfeldt: The AppHeader is built to be used by different applications, and is therefore placed in common. The component SystemBanners is unique for the application which is why it's placed in webcert. No components from webcert should be used in common.

perfli (approved):

PR #321, opened 2022-03-07
INTYGFV-14421: Ensure complements are rendered after all sub questions

perfli:
Business rules:
* Request for complements are only sent for main questions
* Notice on request for complements is shown after the main question and
sub questions if any

Note: the code still supports complements for sub questions, I saw no
need rewrite that logic.


mhornfeldt (approved):

mww-sopra-steria (approved):
I couldn't think of a better name... :-)

PR #322, opened 2022-03-07
Update readme with accurate info

perfli:


README.md
```diff
@@ -1,142 +1,126 @@
 # `Frontend`
 
-## Prerequisite
+## Pre-requisites
+
+- Install Java OpenJDK 11
```
mww-sopra-steria, line 5: The instructions related to java is not for the frontend, but for the other apps. Should we refer to the wiki in sklintyg/common instead? Similar to what you do for the sklintyg/devops?

README.md
```diff
@@ -1,142 +1,126 @@
...
+
 ```javascript
 it('displaying empty value', () => {
   const question = createQuestion({ type: CertificateDataValueType.TEXT })
   const { getByText } = render(<UvText question={question} />)
   getByText(/Ej angivet/i)
 })
 ```
+
 Test with snapshot to verify that changes in the component was intentional.
```
mww-sopra-steria, line 118: We could remove snapshot-testing all together. We did try it out but doesn't find valuable enough.

README.md
```diff
@@ -1,142 +1,126 @@
...
+| GIT_CI_BRANCH         | Yes      | Branch in git repository, ex: `master`                                |
+
+To create a frontend-pipeline using the template, you make sure to first login to OpenShift and then run the following command.
 
-To create a frontend-pipeline using the template, you make sure to first login to OpenShift and then run the following command. 
 ```
 `oc process -f pipelinetemplate-build-frontend.yaml -p APP_NAME=webcert-frontend -p RELEASE_VERSION=2021-2 -p GIT_URL=https://github.com/sklintyg/frontend.git -p GIT_CI_BRANCH=master | oc apply  -f -`
 ```
 
 ## Running storybook
```
mww-sopra-steria, line 78: We could remove the storybook section.

README.md
```diff
@@ -1,142 +1,126 @@
...
+      - Press `ctrl` + `shift` + `p`, type `settings` and open `Preferences: Open Settings (JSON)`
+      - Add the following properties
+        ```json
+        {
+          "editor.defaultFormatter": "esbenp.prettier-vscode",
+          "editor.formatOnSave": true
+        }
+        ```
+
+## Run Webcert backend and frontend client
```
mww-sopra-steria, line 37: Should we have app specific information in this instruction? It is convenient and makes it easier the first time someone follows the instructions, but it could also become obsolete if there are changes to Webcert libraries or dependencies.

mww-sopra-steria (changes requested):

README.md
perfli: I thought about it, but not all instructions for backend are necessary. I felt it would be least confusing this way, but perhaps we should structure the docs for `sklintyg/common` a little better instead?

README.md
perfli: Not sure if I understand you correctly. Which apps are you referring to? And why wouldn't we specify dependencies to frontend in its docs? If dependencies change, so should the docs IMO.
Anyhow, if we remove the instructions relating to Webcert's dependencies we should make sure it's very clear what is needed in Webcert's docs.

mww-sopra-steria: I'm referring to Webcert. We have instructions in the devops-repo how to run and how to build the different apps. Having specific information in the frontend-repo on how we build Webcert and the libraries it depends on (infra, common, refdata) is redundant, though it can be convenient.

README.md
mww-sopra-steria: I get your point about it being less confusing. Keep it as you have written and we can change it when we decide to restructure sklintyg/common wiki.

mww-sopra-steria (approved):

README.md
perfli: Ok, then I vote for including the instructions. Figuring out this part was particularly hard and error prone for me when I started.

PR #323, opened 2022-03-08
INTYGFV-14408: Exported TextInput

mhornfeldt:


perfli (approved):
Why use default exports if we´re re-exporting them as named anyways? I see no benefit to default exports over named exports.

PR #324, opened 2022-03-08
INTYGFV-14408: Removed unused prop

mhornfeldt:


perfli (approved):
I don't think trivial changes like this need a code review. Do we have some way of making these kinds of changes pass through?

mhornfeldt:
In the settings that are now I think all PR need a code review. If we would make it optional then we lose the security that code reviews bring. Even though something seems trivial the code might still include problems that another developer will catch.

PR #327, opened 2022-03-17
INTYGFV-14203: Make links in banner open in new tab and add icon

HelenaEdblad:


packages/common/src/components/utils/SystemBanner.tsx
```diff
@@ -31,11 +32,21 @@ const SystemBanner: React.FC<Props> = ({ banner }) => {
     return 'ic-alert--observe'
   }
 
+  const modifyBannerContent = (content: string) => {
+    let bannerContent = content
+    bannerContent = bannerContent.replace(
+      new RegExp('</a>', 'g'),
+      `<img src=${externalLinkIcon} alt="Länken öppnas i ny flik" style="display: inline; position: relative; top: 3px; left: 2px" /></a>`
```
perfli, line 39: Not a biggie but we could extract the inline style to a styled component or perhaps in a css template literal.

perfli (approved):

packages/common/src/components/utils/SystemBanner.tsx
HelenaEdblad: I know, I did that first but [object Object] was shown instead of the icon. What can I do to avoid the [object Object]?

PR #328, opened 2022-03-18
INTYGFV-14459: Added logout link to header if user has correct resour…

mhornfeldt:
…ce link


mww-sopra-steria (dismissed):
When we do logout, we have to start the logout process by calling the idp. The logout process will redirect the user to the idp and later back to webcert-startpage.

Take a look at how the current client handles logout and get in touch if you want to discuss.

packages/webcert/src/index.css
```diff
@@ -161,4 +161,11 @@ img {
 
 .ic-link-list--nav .ic-button--secondary {
   color: #01A5A3 !important;
+}
+
+.iu-link-divider-left:before {
```
perfli, line 166: Perhaps these styles should reside together with the links in `WebcertHeader` (and not be tied to the css framework class)?

packages/webcert/src/page/SearchAndCreatePage.tsx
```diff
@@ -29,6 +30,12 @@ const CreateCertificatePage: React.FC = () => {
     ReactTooltip.hide()
   }, [patient])
 
+  useEffect(() => {
```
perfli, line 33: Should this kind authorization logic reside in that page components? I foresee a lot of duplication for each page we create. Perhaps this kind of logic should be centralized, e.g. in a parent component or a HOC.

perfli (changes requested):
I think there might be some overlap in this issue and INTYGFV-14452

packages/webcert/src/page/SearchAndCreatePage.tsx
mhornfeldt: Which check are you referring to, the resourceLink or the user object?

perfli: I'm thinking of all the code in the `useEffect`, checking to see if the user is authorized to view a certain page and redirecting based on that. I've also seen some warnings relating to redirecting during rendering of a component, perhaps we should prefer to to redirects with `<Redirect />` instead of `history.push()` in components.

perfli: For an example of what I mean, you could take a look at https://github.com/sklintyg/frontend/blob/feature/INTYGFV-14452_login_R2/packages/webcert/src/utils/withLoggedInUserRedirect.tsx

perfli (changes requested):
Regarding `PageWithRedirect`:
I think this component would be better served as a HOC, especially since with 14452 we're going to have the same kind och logic using two different patterns. Also, I think it would be better named as something like `withResourceAccess` or `withResourceControl`.
Perhaps we should write tests for this logic as well?

mhornfeldt:
> Regarding `PageWithRedirect`:
> 
> I think this component would be better served as a HOC, especially since with 14452 we're going to have the same kind och logic using two different patterns. Also, I think it would be better named as something like `withResourceAccess` or `withResourceControl`.
> 
> Perhaps we should write tests for this logic as well?
> 
> And another thought; I believe it would be better if we also let that component own the resource mapping, instead of passing it as a prop/parameter.

If the HOC component gets the userLinks from the state how will the component know which resourceLinkType to validate?

perfli (approved):

PR #346, opened 2022-04-26
INTYGFV-14464: Fixed ts errors for signed list

mhornfeldt:


packages/webcert/src/store/list/listSelectors.ts
```diff
@@ -14,7 +14,7 @@ export const getActiveListType = (state: RootState): ListType => state.ui.uiList
 
 export const hasListError = (state: RootState): boolean => state.ui.uiList.listError
 
-export const getListTotalCount = (state: RootState): number => state.ui.uiList.totalCount
+export const getListTotalCount = (state: RootState): number | undefined => state.ui.uiList.totalCount
```
perfli, line 17: Why do we need to return `undefined` from selector? I think it would make more sense to return `0` if state is `undefined` instead of forcing all clients to make `undefined` checks.

I believe we should try to avoid returning null/undefined, most of the time it's not necessary and leads to a lot of bloated code.

perfli (approved):

packages/webcert/src/store/list/listSelectors.ts
mhornfeldt: TotalCount is undefined before it has been set to indicate to the list the different states of a completely empty list, an empty filtered list etc.

perfli: I understand. Perhaps that merits a new state selector? It would make the different states more explicit

PR #356, opened 2022-05-06
INTYGFV-14418: Added spinner when questions are loading

mhornfeldt:


mhornfeldt:
Todo: Fix height of container when loading

packages/common/src/components/utils/StatusWithIcon.tsx
```diff
@@ -7,7 +7,8 @@ const Wrapper = styled.div`
   display: flex;
   align-items: center;
 
-  .statusContent > a {
+  .statusContent > a,
```
perfli, line 10: In general I think we should avoid styles that "reach into" child components. Alternatives could be to keep contextual styles within the child component (if they're of a general kind) eg.
```typescript
import ContextualComponent from './ContextualComponent'

const Wrapper = styled.div`
  // styles

  ${ContextualComponent} & {
    // contextual styles
  }
`
```
it makes it easier to understand how the component is used and eases refactoring

packages/webcert/src/store/question/questionReducer.ts
```diff
@@ -59,6 +61,7 @@ const getInitialState = (): QuestionState => {
     isSendingQuestion: false,
     functionDisablers: [],
     errorId: '',
+    isLoadingQuestions: true,
```
perfli, line 64: It a bit counterintuitive to let initial state to be true, it would be better to ensure correct behavior with initial state set to false

perfli (changes requested):

packages/webcert/src/store/question/questionReducer.ts
mhornfeldt: Then I have to have undefined in the beginning which I have gotten comments to avoid. Which is to be preferred out of the two?

perfli: What I'm suggesting is that the model matches the expectation of the data. I'm not sure what the client's code looks like but my thinking is that `isLoadingQuestions` should only be able to be `true` or `false`, and only `true` if questions actually are loading. That might mean that some additional checks need to occur in the client to handle state correctly.

packages/common/src/components/utils/StatusWithIcon.tsx
perfli: The main idea is that the component that needs contextual styles (`<a />`) owns it's own styles. It's kind of like Inversion of Control in OO. In this case it might not matter, or even be applicable, but I wanted to introduce the suggestion as I think it will make the code easier to reason about and refactor.

packages/webcert/src/store/question/questionReducer.ts
mhornfeldt: Fixed this

perfli (approved):

PR #359, opened 2022-05-10
INTYGFV-14621: Added tabs in header for user to navigate between pages

mhornfeldt:


packages/common/src/components/AppHeader/AppHeaderTabs.test.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import '@testing-library/jest-dom'
+import { render, screen } from '@testing-library/react'
+import AppHeaderTabs from './AppHeaderTabs'
+import { MemoryRouter, Route } from 'react-router-dom'
+import { Provider } from 'react-redux'
+import store from '@frontend/webcert/src/store/store'
+
+const getTabs = () => {
```
perfli, line 9: We should add typings to the test

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react'
+import { UserTab } from '../../types/utils'
+import { CustomButton } from '@frontend/common'
+import styled from 'styled-components'
+import { useHistory, useRouteMatch } from 'react-router-dom'
+
+const Wrapper = styled.nav`
+  border: 1.5px solid rgba(1, 165, 163, 0.08);
```
perfli, line 8: Don't we have predefined constants for the colors we use?

packages/webcert/src/App.tsx
```diff
@@ -31,6 +31,7 @@ function App(): JSX.Element {
     window.addEventListener('beforeunload', handleWindowBeforeUnload)
     dispatch(cancelLogout())
     dispatch(getUser())
+    dispatch(getUserTabs())
```
perfli, line 34: Do we need to wait on the data before letting the user interact with the app?

perfli (changes requested):

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
mhornfeldt: We tried using the css function var() before to access the colors. It worked locally but in the test env it didn't find the colors. Is there another way to do it?

perfli: You could just define them in javascript, that's better than reproducing the values all over the code base.

packages/webcert/src/App.tsx
mhornfeldt: No

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
mhornfeldt: Fixed

mhornfeldt: Fixed

perfli (approved):

HelenaEdblad (approved):

packages/webcert/src/utils/userTabsUtils.tsx
```diff
@@ -0,0 +1,73 @@
+import { ResourceLink, ResourceLinkType, UserStatistics, UserTab } from '@frontend/common'
```
perfli, line 1: It's a bit redundant to have "Utils" in the file name `userTabsUtils`

PR #360, opened 2022-05-10
INTYGFV-14628: Reset certificate state when search create page is acc…

mhornfeldt:
…essed


mhornfeldt:
> It is an important behaviour that the state is being reset. Should have a unit test that verifies the behaviour.

I'll add a test! Should I reset some other variables in the certificate state also? The reported bug was only related to validationErrors.

mww-sopra-steria:
> > It is an important behaviour that the state is being reset. Should have a unit test that verifies the behaviour.
> 
> I'll add a test! Should I reset some other variables in the certificate state also? The reported bug was only related to validationErrors.

Thanks for bringing it up! I was a bit quick and actually just took for granted that the resetState already existed. But we didn't have one until now in the certificate-slice.

Take a look at the questionReducer and introduce a resetState that resets the whole state.

PR #361, opened 2022-05-10
INTYGFV-14127: Validation error for dates should always be shown

mhornfeldt:


mhornfeldt:
@mww-sopra-steria I understand this solution is not optimal but I feel like this is the only way to solve this if we don't wanna change the object which also is used in the original application (if the flag would be set when the error message is added in backend). There is no type or id that I can use to identify the validationErrors that should be shown without clicking "Sign". Another solution would maybe be to add this flag before sending the validationError from the server to the client, however this still includes this "ugly" identification using the text. I didn't think it was worth spending a lot of time and changing the logic in the server for two validationErrors with very specific behaviour. What do you think?

mww-sopra-steria (changes requested):
I read your comment and understand the challenge. But the solution is not maintainable and will break as soon as a change is introduced in the text. We have to find a different way to solve it.

mhornfeldt:
> I read your comment and understand the challenge. But the solution is not maintainable and will break as soon as a change is introduced in the text. We have to find a different way to solve it.

What if I would change the type of the validationError to indicate that it is always visible? Then I'll avoid having text specific logic in the frontend and I don't have to change the structure of the original object

mww-sopra-steria:
> > I read your comment and understand the challenge. But the solution is not maintainable and will break as soon as a change is introduced in the text. We have to find a different way to solve it.
> 
> What if I would change the type of the validationError to indicate that it is always visible? Then I'll avoid having text specific logic in the frontend and I don't have to change the structure of the original object

Lets have a chat about this tomorrow morning!

mww-sopra-steria (approved):
Approved! And we agreed that unit tests should be added before merge.

PR #364, opened 2022-05-12
INTYGFV-14624: Modals in certificate should be placed over whole page

mhornfeldt:


perfli (changes requested):
This solution seems a bit off in my opinion (adding a prop that needs to be passed up/down the component hierarchy in order to affect layer ordering). Modals are by definition placed on top of current content so I think a central/global handling of modals will be a better fit.
One might event consider a consolidation with the error modals.

PR #367, opened 2022-05-13
INTYGFV-14453: Choose care provider modal

HelenaEdblad:


packages/webcert/src/App.tsx
```diff
@@ -52,6 +53,7 @@ function App(): JSX.Element {
       <BrowserRouter>
         <ErrorBoundary fallbackRender={({ error }) => <>Ett fel har inträffat: {error.message}</>} onError={onError}>
           <ErrorComponent />
+          <CareProviderModal title="Välj vårdenhet" />
```
mhornfeldt, line 56: The text could be set in the component instead of App

packages/webcert/src/components/modal/CareProviderModal.test.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+import { Provider } from 'react-redux'
+import { BrowserRouter } from 'react-router-dom'
+import store from '../../store/store'
+import CareProviderModal from './CareProviderModal'
+import { updateUser } from '../../store/user/userActions'
+import { LoginMethod, SigningMethod, Unit, User } from '@frontend/common'
+
+describe('Care provider modal', () => {
```
mhornfeldt, line 10: More tests should be added either in this component or the child components. For example test that the units are printed, that the show/hide units functionality works etc.

packages/webcert/src/components/modal/CareProviderModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+      setIsOpen(true)
+    }
+  }, [user, dispatch])
+
+  if (!isOpen || user?.careProviders.length === 0) {
+    return null
+  }
+
+  const { careProviders } = user as User
```
mhornfeldt, line 50: What is the functionality of this?

packages/webcert/src/components/modal/CareProviderModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const [isOpen, setIsOpen] = useState(false)
+
+  const handleChooseUnit = (event: React.MouseEvent) => {
+    const unitId = event.currentTarget.id
+
+    dispatch(setUnit(unitId))
+    setIsOpen(false)
+  }
+
+  useEffect(() => {
```
mhornfeldt, line 36: Is the local variable open necessary or can the actual check if the unit is set be used instead?

packages/webcert/src/components/modal/CareProviderModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }, [user, dispatch])
+
+  if (!isOpen || user?.careProviders.length === 0) {
+    return null
+  }
+
+  const { careProviders } = user as User
+
+  return (
+    <FocusTrap active={isOpen}>
```
mhornfeldt, line 53: Can the toast modal be used instead and deactivate the automatic closing? Otherwise the modal can be extracted into a more generalized component (and placed in common) and the content be sent in as children.

packages/webcert/src/components/modal/CareProviders.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface Props {
+  careProviders: CareProvider[]
+  chooseUnit: (event: React.MouseEvent) => void
+}
+
+export const CareProviders: React.FC<Props> = ({ chooseUnit, careProviders }) => {
+  return (
+    <>
+      {careProviders.map((careProvider) => (
+        <details className="ic-card ic-card--expandable ic-card--sm-unset-style ic-expandable" key={careProvider.id} open>
```
mhornfeldt, line 14: Can the expandable text be extracted in a more generalized component similar to the modal?

packages/webcert/src/store/user/userMiddleware.test.ts
```diff
@@ -42,4 +42,13 @@ describe('Test certificate middleware', () => {
       expect(fakeAxios.history.get.length).toBe(0)
     })
   })
+
+  describe('Handle set unit', () => {
```
mhornfeldt, line 46: More tests should be added to middleware. For example to test the success-function after setting unit and to test the url for the api call

packages/webcert/src/components/modal/ExpandableCareUnit.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { useEffect, useState } from 'react'
```
mhornfeldt, line 1: Also here it feels like the expandable components can be extracted to general, reusable components and separated from the logic of the Units

mhornfeldt (changes requested):
I've left some comments about maybe extracting the expandable text-functionality and modal in more general components, these are just suggestions. Other than that there are some missing tests and minor changes.

packages/webcert/src/components/modal/CareProviderModal.tsx
HelenaEdblad: This was removed when I restructured the code.

packages/webcert/src/components/modal/CareProviderModal.tsx
HelenaEdblad: A new generic popup modal has been created.

packages/webcert/src/feature/careProvider/CareProviderModalContent.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const handleChooseUnit = (event: React.MouseEvent) => {
+    const unitId = event.currentTarget.id
+
+    dispatch(setUnit(unitId))
+  }
+
+  return (
+    <>
+      {user?.careProviders.map((careProvider, idx) => {
+        const headings = [careProvider.name, 'Ej hanterade ärenden', 'Ej signerade utkast']
```
mhornfeldt, line 21: Should these headings be generated from which pages the user can access? In that case the texts can be taken from the resource links

packages/webcert/src/feature/careProvider/ExpandableTableRow.tsx
```diff
@@ -0,0 +1,88 @@
...
+`
+
+interface ExpandedUnitProps {
+  isExpanded: boolean
+  unit: Unit
+  handleChooseUnit: (event: React.MouseEvent) => void
+  statistics: UnitStatistic
+}
+
+const ExpandedUnit: React.FC<ExpandedUnitProps> = ({ unit, isExpanded, handleChooseUnit, statistics }) => {
```
mhornfeldt, line 20: This can be extracted to its own file. Also I think a lot of functionality in this file can be made more generalized. The name of the component is ExpandableTableRow but it has a close connection to the units. That content can be expandable is something that has been used before in the application and will be used in the future. Therefore you could see if you can use an existing component to create the expandable content, or create a new generalized component. I think it would be good to separate displaying the unit information and the exapandable logic so that the expandable component can be used without the unit and the unit information can be shown without being expandable. Maybe you could even separate the unit and the unit statistics into two components.

packages/webcert/src/feature/careProvider/ExpandableTableRow.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <td>
+          {units.length > 0 && (
+            <StyledArrow icon={isExpanded ? faAngleUp : faAngleDown} className="iu-color-cta-dark iu-mr-300" onClick={toggleOpen} />
+          )}
+          <button className="ic-link iu-text-left" type="button" id={careUnitId} onClick={handleChooseUnit}>
+            {careUnit}
+          </button>
+        </td>
+        <td>
+          {questionsOnUnit} {units.length > 0 && `(total ${questionsOnUnit + questionsOnSubUnits})`}
```
mhornfeldt, line 70: This could be extracted to a function so you don't have to duplicate the code for two different variables (line 73)

packages/webcert/src/store/user/userMiddleware.test.ts
HelenaEdblad: Is it enough with the test in CareProviderModal.test?

mhornfeldt: This comment still holds and now also applies for the statistics

mhornfeldt (changes requested):

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+  it('should show care units when care provider modal is open', () => {
+    testStore.dispatch(updateUser(getUserWithEmptyUnit()))
+    testStore.dispatch(updateUserStatistics(getUserStatistics()))
+
+    renderComponent()
+    expect(screen.getByText('Care unit')).toBeInTheDocument()
+  })
+})
+
+const getUserWithEmptyUnit = (): User => {
```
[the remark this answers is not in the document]
HelenaEdblad: Yes, or we do it when the file starts getting bigger than what it is now?

packages/webcert/src/feature/careProvider/CareProviderModalContent.tsx
HelenaEdblad: According to the requirements the columns for "Ej hanterade ärenden" and "Ej signerade utkast" is not dependent on what the user can access. I will leave them hard coded for now and we can change it if the requirements change.

packages/webcert/src/store/user/userMiddleware.ts
```diff
@@ -142,8 +146,30 @@ const handleGetUserStatistics: Middleware<Dispatch> = ({ dispatch }: MiddlewareA
...
+}
+
+const handleSetUnit: Middleware<Dispatch> = ({ dispatch }: MiddlewareAPI) => () => (action: AnyAction): void => {
+  dispatch(
+    apiCallBegan({
+      url: `/api/user/unit/${action.payload}`,
+      method: 'POST',
+      onStart: setUnitStarted.type,
+      onSuccess: setUnitSuccess.type,
+      onError: apiSilentGenericError.type,
```
mhornfeldt, line 165: Should this result in a silent error?

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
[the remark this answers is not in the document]
mhornfeldt: There's already three functions in this test class that could be used by other tests so I think it's the right time to extract them

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
```diff
@@ -0,0 +1,204 @@
...
+    testStore.dispatch(updateUser(getUser()))
+
+    renderComponent()
+    expect(screen.queryByRole('dialog')).not.toBeInTheDocument()
+  })
+
+  it('should close modal when unit is chosen', async () => {
+    testStore.dispatch(updateUser(getUserWithEmptyUnit()))
+    testStore.dispatch(updateUserStatistics(getUserStatistics()))
+    fakeAxios.onPost('/api/user/unit/1234a').reply(200, { user: getUser() })
```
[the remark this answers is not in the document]
HelenaEdblad: Because we wanted to test the click on a unit in the modal and we needed the onPost to update the store with new data. We could have updated the store directly but then the click would have been unecessary.

packages/common/src/components/Table/ExpandableTableRow.tsx
```diff
@@ -0,0 +1,46 @@
+import { faAngleDown, faAngleUp } from '@fortawesome/free-solid-svg-icons'
```
mhornfeldt, line 1: Tests are missing for ExpandableTableRow

packages/webcert/src/feature/careProvider/CareProviderModalContent.tsx
```diff
@@ -0,0 +1,86 @@
...
+    })
+  }
+
+  const handleChooseUnit = (event: React.MouseEvent) => {
+    const unitId = event.currentTarget.id
+
+    dispatch(setUnit(unitId))
+  }
+
+  return (
```
mhornfeldt, line 44: This return statement is long and kind of hard to read. Could you maybe extract some code in functions to make it easier to read?

mhornfeldt (changes requested):

packages/common/src/components/utils/Modal/PopUpModal.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const PopUpModal: React.FC<Props> = ({ open, modalTitle, children }) => {
+  if (!open) {
+    return null
+  }
+
+  const handleClose = () => {
+    return null
+  }
```
perfli, line 16: Why return `null`? If it's a no-op the perhaps this would be clearer
```suggestion
  const handleClose = () => {}
```

perfli (changes requested):

packages/webcert/src/feature/careProvider/CareProviderModalContent.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export const CareProviderModalContent: React.FC = () => {
+  const dispatch = useDispatch()
+  const user = useSelector(getUser)
+  const unitStatistics = useSelector(selectUnitStatistics)
+
+  const getUnitStatisticsLiteral = (amountOnUnit: number, amountOnOtherUnit?: number, careUnit?: CareUnit) => {
+    const showTotal = careUnit && careUnit?.units.length > 0
+    return `${amountOnUnit} ${showTotal && amountOnOtherUnit !== undefined ? `(total ${amountOnUnit + amountOnOtherUnit})` : ''}`
+  }
```
perfli, line 15:
```suggestion
  const getUnitStatisticsLiteral = (amountOnUnit: number, amountOnOtherUnit = 0, careUnit?: CareUnit) => {
    const showTotal = careUnit && careUnit?.units.length > 0 && amountOnOtherUnit > 0
    return `${amountOnUnit} ${showTotal ? `(total ${amountOnUnit + amountOnOtherUnit})` : ''}`
  }
```

perfli (approved):

mhornfeldt (approved):

PR #373, opened 2022-05-20
INTYGFV-14621: Updates to user tabs

mhornfeldt:


packages/common/src/components/AppHeader/AppHeaderTabs.tsx
```diff
@@ -1,15 +1,23 @@
...
+import { useDispatch } from 'react-redux'
+import { resetListState } from '@frontend/webcert/src/store/list/listActions'
 
 const Wrapper = styled.nav`
   button {
     display: inline;
     margin-right: 48px;
   }
+
+  .ic-topnav__link {
```
perfli, line 18: We should couple our styles to the css framework, it's better to create our own class name or style component

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
```diff
@@ -29,7 +46,15 @@ const AppHeaderTabs: React.FC<Props> = ({ tabs }) => {
   }
 
   const handleClick = (tab: UserTab) => {
-    history.push(tab.url)
+    if (match.url !== tab.url) {
+      dispatch(resetPatientState())
```
perfli, line 50: Why do we reset patient state between navigations?

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
```diff
@@ -39,9 +64,10 @@ const AppHeaderTabs: React.FC<Props> = ({ tabs }) => {
           <a
             tabIndex={0}
```
perfli, line 65: Why do we have `tabIndex={0}` on a link? The should be "navigatable" by default

perfli (approved):

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
mhornfeldt: Because otherwise the searched patient will be saved if the user goes from the create page to a list. This is not how it should work. Also, if the user has searched for a patient and then clicks on the Search/create-tab the search page should be showed, not the patient page.

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
mhornfeldt: Maybe it has something to do with the UX-components used but without the tabIndex the tabs are not focusable.

packages/common/src/components/AppHeader/AppHeaderTabs.tsx
perfli: I get that we don't want to show the previously searched patient when user navigates to search page again, but don't we want to keep the patient while navigating between lists an certificates?
Also, I realized we get some weird couplings when we reference webcert specific actions from a component in common, we should invert that coupling.

mhornfeldt: You can try it yourself in the test environments but when I navigate: Search with patient -> List -> Search, the patient is reset for the original application

perfli: Never mind, I thought we needed the patient info on the list views.
Though I pulled the code and tried it out locally and there's a lot of duplicated api calls. I doesn't look like they bring any new data and could probably be cached.
image
Also, when I switch to Signed Certificates tab, I get an error while retrieving the list (probably because I haven't backend) and I end up in an infinite loop of api calls. That shouldn't be able to happen.
image

PR #381, opened 2022-05-24
TYGFV-14465: Fixes to signed certificates list

mhornfeldt:


packages/webcert/src/feature/list/filter/DateRangeFilter.tsx
```diff
@@ -146,7 +159,7 @@ const DateRangeFilter: React.FC<Props> = ({ config, onChange }) => {
             max={configFrom.max}
             min={configFrom.min}
           />
-          {fromValidationError && (
+          {!validationError && fromValidationError && (
```
perfli, line 162: At the first look I was a bit confused about the wording of the logic. Perhaps it would be clearer if `validationError` was renamed to something like `globalValidationError`

perfli (approved):

PR #384, opened 2022-05-25
INTYGFV-14460: Re-implement person id validation from old client

perfli:


packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
-export const formatPersonId = (id: string) => {
+export const formatPersonId = (id: string): string => {
   let cleanId = id.replace(/\D/g, '')
   if (cleanId.length > 8) {
     cleanId = cleanId.slice(0, 8) + '-' + cleanId.slice(8, cleanId.length)
   }
   return cleanId
 }
+
+const isPersonnummerValid = (value: string): boolean => {
```
mhornfeldt, line 15: We should avoid swedish words in the code. PersonId and CoordinationNumber could be used instead.

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+  return isCheckDigitValid(parts[2] + parts[3] + parts[4] + parts[6])
+}
+
+const isSamordningsnummerValid = (value: string): boolean => {
+  const SAMORDNINGSNUMMER_REGEXP = /^(\d{2})(\d{2})(\d{2})([6-9]\d)-?(\d{4})$/
+  const parts = SAMORDNINGSNUMMER_REGEXP.exec(value)
+
+  if (!parts) return false
+
+  const day = parseInt(parts[4], 10) - 60 // 60 is the special number for samordningsnummer.
```
mhornfeldt, line 36: We should avoid comments. Use a variable name for 60 instead giving it a name where the contents of the comment is understood

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+
+  if (!isValid(parsedDate) || isFuture(parsedDate) || !isDateInValidRange(parsedDate)) {
+    return false
+  }
+
+  return isCheckDigitValid(parts[2] + parts[3] + parts[4] + parts[5])
+}
+
+const isCheckDigitValid = (value: string): boolean => {
+  // Multiply each of the digits (except the control digit) with 2,1,2,1,...
```
mhornfeldt, line 47: Replace comment with variable/function or similar, or simply remove it

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+    .join('')
+
+  const sum = multipliedDigits.split('').reduce((prev: number, curr: string) => prev + parseInt(curr, 10), 0)
+  const checkDigit = parseInt(value.substring(value.length - 1), 10)
+
+  return (10 - (sum % 10)) % 10 === checkDigit
+}
+
+const isDateInValidRange = (start: Date): boolean => {
+  // A date is not allowed to be greater than 125 years
```
mhornfeldt, line 64: Same as above

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+
+const isPersonnummerValid = (value: string): boolean => {
+  const PERSONNUMMER_REGEXP = /^(\d{2})(\d{2})(\d{2})([0-3]\d)([-+]?)(\d{4})$/
+  const parts = PERSONNUMMER_REGEXP.exec(value)
+
+  if (!parts) return false
+
+  const parsedDate = parseISO(parts[1] + parts[2] + '-' + parts[3] + '-' + parts[4])
+
+  if (!isValid(parsedDate) || isFuture(parsedDate) || !isDateInValidRange(parsedDate)) {
```
mhornfeldt, line 23: Maybe isDateValid isFutureDate follows the naming pattern more since you have included date on the last one

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+    return false
+  }
+
+  return isCheckDigitValid(parts[2] + parts[3] + parts[4] + parts[6])
```
mhornfeldt, line 27: Since the parts are used in several places could the algorithm be easier to read if we saved them in variables with explainable names? Right now it's hard to understand what is in each part. Same goes for coordination number

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+const isSamordningsnummerValid = (value: string): boolean => {
+  const SAMORDNINGSNUMMER_REGEXP = /^(\d{2})(\d{2})(\d{2})([6-9]\d)-?(\d{4})$/
+  const parts = SAMORDNINGSNUMMER_REGEXP.exec(value)
+
+  if (!parts) return false
+
+  const day = parseInt(parts[4], 10) - 60 // 60 is the special number for samordningsnummer.
+  const parsedDate = parseISO(parts[1] + parts[2] + '-' + parts[3] + '-' + pad(day))
+
+  if (!isValid(parsedDate) || isFuture(parsedDate) || !isDateInValidRange(parsedDate)) {
```
mhornfeldt, line 39: The same check is done for person id, could it be extracted to a common function to avoid duplicating code?

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+  const multipliedDigits = value
+    .substring(0, value.length - 1)
+    .split('')
+    .map((digit: string, idx: number) => {
+      const multiplier = idx % 2 === 0 ? 2 : 1
+      return parseInt(digit, 10) * multiplier
+    })
+    .join('')
+
+  const sum = multipliedDigits.split('').reduce((prev: number, curr: string) => prev + parseInt(curr, 10), 0)
```
mhornfeldt, line 57: Consider naming the variables previous and current instead to increase readability

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+const isDateInValidRange = (start: Date): boolean => {
+  // A date is not allowed to be greater than 125 years
+  const durationFromToday = intervalToDuration({
+    start,
+    end: new Date(),
+  })
+  return durationFromToday.years ? durationFromToday.years < MAXIMUM_AGE + 1 : false
+}
+
+const pad = (number: number): string => (number < 10 ? '0' + number : number.toString())
```
mhornfeldt, line 72: Not clear from the name what this does, could padWithZero be a better name?

packages/common/src/utils/personIdValidatorUtils.ts
```diff
@@ -1,29 +1,72 @@
...
+  const parsedDate = parseISO(parts[1] + parts[2] + '-' + parts[3] + '-' + pad(day))
+
+  if (!isValid(parsedDate) || isFuture(parsedDate) || !isDateInValidRange(parsedDate)) {
+    return false
+  }
+
+  return isCheckDigitValid(parts[2] + parts[3] + parts[4] + parts[5])
+}
+
+const isCheckDigitValid = (value: string): boolean => {
```
mhornfeldt, line 46: This function is hard to read which is understandable since it's a lot of calculations. However, if it could be split into two functions, one for calculating the check digit and one for checking, it could be easier to refactor the function in the future.

mhornfeldt (changes requested):

packages/common/src/utils/personIdValidatorUtils.ts
perfli: I agree, we should avoid comments, but not all comments. I think introducing a constant for 60 would detract from the ability to understand the code, especially since there's no name for the constant.

packages/common/src/utils/personIdValidatorUtils.ts
perfli: This one is not useful, I'll remove it

packages/common/src/utils/personIdValidatorUtils.ts
perfli: In that case maybe I should remove "Date" from the last one since the other two come from the date-fns library

packages/common/src/utils/personIdValidatorUtils.ts
perfli: I've actually already done that since I needed to calculate a valid check digit for the tests.

mhornfeldt (approved):

PR #385, opened 2022-05-31
INTYGFV-14679: Saved validation errors for list globally

mhornfeldt:


mhornfeldt:
> When we have a global state for lists and three different list views, how do we ensure the state for one view doesn't bleed over to the other views? As I mentioned earlier I saw one such behavior related to error handling.

The state for the list is reset when switching list.

PR #386, opened 2022-06-09
INTYGFV: 14453: Add resource link for choose unit

HelenaEdblad:


packages/common/src/components/Footer/FooterCookiesModal.tsx
```diff
@@ -14,8 +14,17 @@ const ExternalLinkIconImg = styled.img`
   position: relative;
 `
 
+const StyledTextWithInfoModal = styled(TextWithInfoModal)`
```
mhornfeldt, line 17: How is this change realted to this Jira?

packages/common/src/components/utils/Modal/TextWithInfoModal.tsx
```diff
@@ -16,10 +16,9 @@ interface Props {
   text: string
   modalTitle: string
   className?: string
-  additionalStyles?: string
 }
 
-const TextWithInfoModal: React.FC<Props> = ({ text, modalTitle, children, additionalStyles, className }) => {
```
mhornfeldt, line 22: This can't be removed since the component is used in other places in the code which uses this attribute. If you feel the need to remove it you need to refactor the other components.

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -1,15 +1,17 @@
...
 import { CareProviderModalContent } from './CareProviderModalContent'
 
 const CareProviderModal: React.FC = () => {
   const user = useSelector(getUser)
   const isLoadingUserStatistics = useSelector(selectIsLoadingUserStatistics, _.isEqual)
+  const userLinks = useSelector(getUserResourceLinks)
+  const chooseUnitLink = userLinks?.find((link) => link.type === ResourceLinkType.CHOOSE_UNIT)
 
-  const showCareProviderModal = !!user && !user?.loggedInUnit.unitId
+  const showCareProviderModal = !user?.loggedInUnit.unitId && !!chooseUnitLink
```
mhornfeldt, line 14: If you are checking for the resource link shouldn't the logic of checking if the unit has been set in the user object be moved to backend? If the modal should not be shown the resource link should either be missing or disabled.

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
```diff
@@ -49,6 +50,7 @@ describe('Care provider modal', () => {
   it('should show care provider modal if logged in unit is not set', () => {
```
mhornfeldt, line 50: Add test that if the resource link is not avaialble the modal is not shown

mhornfeldt (changes requested):

packages/common/src/components/Footer/FooterCookiesModal.tsx
HelenaEdblad: It's not, but the arrow used to be there before and has somehow disappeared. So I thought the quickest way was to just put it back again, instead of creating a new ticket in Jira etc.

packages/common/src/components/utils/Modal/TextWithInfoModal.tsx
HelenaEdblad: I removed it since it is not used anywhere in the component. It's just imported as a prop and then not used. I had a quick look throw where the component is used and additionalStyles was only used in one place so I refactored that code to be className instead.

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
HelenaEdblad: Yes, I agree. I need help with that though. I don't know how to code that in backend.

packages/common/src/components/utils/Modal/TextWithInfoModal.tsx
mhornfeldt: It looks like the reason why it is not used in the component is because it wrongly removed when the component was refactored from a link to a button. When we change code which is not realted to the Jira we need to make sure that we don't remove functionality. Otherwise we will create hidden regression bugs since test won't look at the code which is not related to the Jira.

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
mhornfeldt: For this PR the logic needs to be removed and tests needs to be written reflecting that the frontend will only check if the resource link is active/available.

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
```diff
@@ -64,6 +74,8 @@ describe('Care provider modal', () => {
   it('should close modal when unit is chosen', async () => {
```
mhornfeldt, line 74: Here it seems like you are referring to backend logic in your frontend test. The test itself might not be wrong but from my point of view it doesn't seem like the text explains what the test is actually testing. Isn't what you actually want to test if the user chooses a unit then certain api-calls are made (api-call to choose unit, and also maybe to update user and to update the user resource links)? And another test which checks if these api-calls results in the resource link for choose unit to be missing then the modal closes?

mhornfeldt (dismissed):

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
HelenaEdblad: The name of the test is correct because I'm only testing the user interaction. The other tests that you talk about are among the user middleware tests.

perfli (approved):

PR #393, opened 2022-06-14
INTYGFV-14454: Change unit

HelenaEdblad:


packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -1,21 +1,50 @@
...
+import { useDispatch, useSelector } from 'react-redux'
+import { updateIsCareProviderModalOpen } from '../../store/user/userActions'
+import { getUser, selectIsLoadingUserStatistics, getIsCareProviderModalOpen, getUserResourceLinks } from '../../store/user/userSelectors'
 import { CareProviderModalContent } from './CareProviderModalContent'
 
 const CareProviderModal: React.FC = () => {
+  const dispatch = useDispatch()
+  const user = useSelector(getUser)
   const isLoadingUserStatistics = useSelector(selectIsLoadingUserStatistics, _.isEqual)
+  const isCareProviderModalOpen = useSelector(getIsCareProviderModalOpen)
```
mhornfeldt, line 13: Is it necessary to save this variable in the state? Why can't a local variable be used?

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -1,21 +1,50 @@
...
+      return chooseUnitLink.name
+    } else if (changeUnitLink) {
+      return changeUnitLink.name
+    } else {
+      return ''
+    }
+  }
+
+  useEffect(() => {
+    if (chooseUnitLink && !user?.loggedInUnit.unitId) {
```
mhornfeldt, line 29: Why is the logic to check if the unit id is not set necessary? Shouldn't it be enough with the resource link? If logical checks are needed these should be put in the backend with the resource link

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -1,21 +1,50 @@
...
     return null
   }
 
   return (
-    <PopUpModal modalTitle="Välj vårdgivare" open={!!chooseUnitLink}>
+    <PopUpModal
+      modalTitle={getModalTitle()}
+      open={isCareProviderModalOpen}
+      handleClose={handleClose}
+      showCloseButton={!!user?.loggedInUnit.unitId}>
```
mhornfeldt, line 47: This logic seems to be related to if the resource link choose unit is available

mhornfeldt: Tests are missing for this logic

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
```diff
@@ -63,13 +63,13 @@ describe('Care provider modal', () => {
...
 
-  describe('Tests with common setup', () => {
+  describe('Tests with no logged in unit', () => {
     beforeEach(() => {
       testStore.dispatch(updateUser(getUserWithEmptyUnit()))
       testStore.dispatch(updateUserStatistics(getUserStatistics()))
     })
 
-    it('should show care provider modal if choose unit resource link exists', () => {
+    it('should show care provider modal if choose unit resource link exists and logged in unit is not set', () => {
```
mhornfeldt, line 72: Shouldn't the logic check "if logged in unit is not set" belong in the backend code?

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
```diff
@@ -24,25 +36,104 @@ const InactiveUnit = styled.span`
...
+  const userLinks = useSelector(getUserResourceLinks)
+  const [isExpanded, setIsExpanded] = useState(false)
+
+  const changeUnitLink = userLinks?.find((link) => link.type === ResourceLinkType.CHANGE_UNIT)
+
+  const toggleMenu = () => {
+    setIsExpanded(!isExpanded)
+  }
+
+  const expandButton = () => {
```
mhornfeldt, line 86: This should be a general component. Maybe there is a component that you could have reused.

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
```diff
@@ -24,25 +36,104 @@ const InactiveUnit = styled.span`
...
+            <InactiveUnit
+              className="iu-ml-400"
+              data-tip="Enheten är markerad som inaktiv i journalsystemet, vilket innebär att viss funktionalitet ej är tillgänglig.">
+              <AlertCircle />
+              <span>Inaktiv enhet</span>
+            </InactiveUnit>
+          ) : null}
+          {expandButton()}
+        </Wrapper>
+        {isExpanded && (
```
mhornfeldt, line 129: This should be extracted to its own component and the expanded button should be made general (see comment above)

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
```diff
@@ -24,25 +36,104 @@ const InactiveUnit = styled.span`
...
-          </InactiveUnit>
-        ) : null}
-      </Wrapper>
+      <RelativeDiv>
+        <Wrapper>
+          <span>
+            {user.loggedInCareProvider.unitName} - {user.loggedInUnit.unitName}
+            <br />
+            <Italic>
+              {changeUnitLink &&
```
mhornfeldt, line 115: It seems like this component is growing, maybe it's a good idea to extract this to its own component

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
mhornfeldt: Tests are missing for closing button, can only find test to open

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
```diff
@@ -1,10 +1,22 @@
...
+import { AppHeaderUserUnit, ResourceLinkType } from '@frontend/common'
+import { getTotalDraftsAndUnhandledQuestionsOnOtherUnits, getUser, getUserResourceLinks } from '../../store/user/userSelectors'
+import { shallowEqual, useDispatch, useSelector } from 'react-redux'
 import { User } from '@frontend/common/src'
 import styled from 'styled-components'
 import AlertCircle from '@frontend/common/src/images/AlertCircle'
+import arrow from '@frontend/common/src/images/arrow-down.svg'
+import { updateIsCareProviderModalOpen } from '../../store/user/userActions'
+
+const ArrowDown = styled.img`
```
mhornfeldt, line 11: This feels too general to belong in this specific component. Also, was there nothing in the UX framwork that could be used?

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
```diff
@@ -24,25 +36,104 @@ const InactiveUnit = styled.span`
...
 
+const StyledButton = styled.button`
+  border: none;
+`
+
+const Link = styled.button`
+  text-decoration: none;
+`
+
+const ExpandedDiv = styled.div`
```
mhornfeldt, line 47: Style should be placed somewhere where it can be reused

packages/webcert/src/components/header/WebcertHeaderUnit.test.tsx
```diff
@@ -1,49 +1,81 @@
...
+      reducer,
+      middleware: (getDefaultMiddleware) => getDefaultMiddleware().prepend(dispatchHelperMiddleware, apiMiddleware, userMiddleware),
+    })
+  })
+
+  afterEach(() => {
+    clearDispatchedActions()
+  })
+
   it('displays care provider and unit that user is logged into', (): void => {
```
mhornfeldt, line 36: It's a bit hard to read but it looks like the test to check if the care provider text shows has disappeared

packages/common/src/components/utils/Modal/PopUpModal.tsx
```diff
@@ -1,19 +1,28 @@
...
 
 interface Props {
   open: boolean
   modalTitle: string
+  handleClose: () => void
+  showCloseButton: boolean
 }
 
-const PopUpModal: React.FC<Props> = ({ open, modalTitle, children }) => {
+const PopUpModal: React.FC<Props> = ({ open, modalTitle, handleClose, showCloseButton, children }) => {
```
mhornfeldt, line 12: Tests are missing for new functionality. Also, now that the PopUpModal can have buttons I don't see the difference between the ModalBase. It seems like this component is just wrapping the ModalBase.

mhornfeldt (changes requested):
When I reviewed choose unit there was functionality that belongs in the backend that was removed from the frontend. However, now it seems to have appeared in the frontend again

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
HelenaEdblad: Since it's just one row I won't extract it to its own component. Or maybe I missunderstand what you mean?

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
HelenaEdblad: The style has been moved to ExpandableBox.

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
HelenaEdblad: I did not find anything in the UX framework I could use. Please let me know if you know a better option. But now I have extracted the arrow into its own component.

mhornfeldt: I think you should try to use this one: https://css.inera.se/?path=/story/components-expandables--card. Maybe not all classNames but some. The solution you have done seems to include som styles that are used in the UX-framwork and your component is missing the aria-attributes for an expandable component so it won't be WCAG compliant. If you use the UX-component then you might get this behavior automatically.

packages/common/src/components/Table/ExpandableTableRow.tsx
```diff
@@ -32,11 +21,7 @@ const ExpandableTableRow: React.FC<Props> = ({ rowContent, id, handleClick, chil
           if (idx === 0) {
             return (
               <td key={idx}>
-                {isExpanded ? (
-                  <ArrowUp src={arrow} alt="" onClick={toggleExpand} className="iu-mr-300" data-testid="expandArrow" />
-                ) : (
-                  <ArrowDown src={arrow} alt="" onClick={toggleExpand} className="iu-mr-300" data-testid="expandArrow" />
-                )}
+                <ArrowToggle onClick={handleToggle} className="iu-mr-200" label="Fäll ut/in tabellrader" isUp={isExpanded} />
```
perfli, line 24: I think you should change the label "Fäll ut/in tabellrader" since it refers to the semantic structure. Perhaps "Fäll ut/in innehåll" might be more appropriate? You might actually have that as a standard label in `ArrowToggle` if no label prop is passed to it.

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -1,23 +1,53 @@
...
+  const dispatch = useDispatch()
   const isLoadingUserStatistics = useSelector(selectIsLoadingUserStatistics, _.isEqual)
+  const isCareProviderModalOpen = useSelector(getIsCareProviderModalOpen)
   const userLinks = useSelector(getUserResourceLinks)
+
   const chooseUnitLink = userLinks?.find((link) => link.type === ResourceLinkType.CHOOSE_UNIT)
+  const changeUnitLink = userLinks?.find((link) => link.type === ResourceLinkType.CHANGE_UNIT)
+
+  const getModalTitle = () => {
+    if (chooseUnitLink) {
```
perfli, line 19: It seems a bit weird that the purpose (title) of the modal is determined by the presence of resource links. Am I to understand that once a unit is chosen `ResourceLinkType.CHOOSE_UNIT` is no longer present?

perfli (approved):

packages/webcert/src/components/header/WebcertHeaderUnit.tsx
HelenaEdblad: Since I will have to change color and size of the arrow I don't think it's worth changing the code. Instead I added aria attributes to the ArrowToggle.

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
HelenaEdblad: Correct, when the user have chosen a unit or if a unit is already set, ResourceLinkType.CHOOSE_UNIT is no longer present.

PR #394, opened 2022-06-21
INTYGFV-14453: Make care provider modal wider

HelenaEdblad:


packages/common/src/components/utils/Modal/ModalBase.tsx
```diff
@@ -36,7 +33,13 @@ const ModalBase: React.FC<Props> = ({ open, handleClose, title, buttons, content
       <FocusTrap active={open}>
         <div tabIndex={0}>
           <Backdrop className="ic-backdrop iu-lh-body" onClick={handleClose} />
-          <WrapText role="dialog" className="ic-modal" aria-labelledby="dialog-title" aria-modal="true">
+          <div
```
perfli, line 36: Why not pass styles via `className` instead and keep element as style component?

HelenaEdblad: I changed it so that a prop desides if its a large modal or not. Let me know if the solutions was too complicated, or can be done in a better way.

packages/common/src/components/utils/Modal/ModalBase.tsx
```diff
@@ -24,9 +28,42 @@ interface Props {
...
+      <div className="ic-modal__head" id="demo-modal-content">
+        <h3 id="dialog-title">{title}</h3>
+      </div>
+      <ModalContentWrapper className="ic-modal__body ic-text">{content}</ModalContentWrapper>
+      <div className="ic-button-group ic-button-group--right">{buttons}</div>
+    </>
+  )
+
+  const getText = () =>
+    largeWidth ? (
```
mhornfeldt, line 57: Instead of doing this you could send the prop for large width to the styled component and apply the style there. By doing this you don't have to change the component modalbase more than sending a prop to WrapText.

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -2,10 +2,15 @@ import { CustomButton, ModalBase, ResourceLinkType } from '@frontend/common'
 import _ from 'lodash'
 import React, { useEffect } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
+import { css } from 'styled-components'
 import { updateIsCareProviderModalOpen } from '../../store/user/userActions'
 import { selectIsLoadingUserStatistics, getIsCareProviderModalOpen, getUserResourceLinks } from '../../store/user/userSelectors'
 import { CareProviderModalContent } from './CareProviderModalContent'
 
+const ModalStyle = css`
```
mhornfeldt, line 10: This is not used from what I can see

packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -47,6 +52,7 @@ const CareProviderModal: React.FC = () => {
       content={<CareProviderModalContent />}
       buttons={!chooseUnitLink && <CustomButton onClick={handleClose} buttonStyle="secondary" text="Avbryt" />}
       enableCross={!chooseUnitLink}
+      largeWidth={true}
```
mhornfeldt, line 55: You don't have to write = true since it's a boolean

mhornfeldt (changes requested):

perfli (approved):

mhornfeldt (approved):

PR #396, opened 2022-06-23
INTYGFV-14614: Show patient status based on new props

perfli:


packages/common/src/types/patient.ts
```diff
@@ -20,6 +20,8 @@ export interface Patient {
   deceased: boolean
   differentNameFromEHR: boolean
   personIdUpdated: boolean
```
mhornfeldt, line 22: Should be removed to match backend

mhornfeldt (approved):

PR #407, opened 2022-07-07
INTYGFV-13881: Fixes to implementation of forward certificate

mhornfeldt:


packages/webcert/src/feature/certificate/Buttons/ForwardCertificateButton.tsx
```diff
@@ -53,7 +55,7 @@ const ForwardCertificateButton: React.FC<Props> = ({
       name={name}
       startIcon={<FontAwesomeIcon size="lg" icon={faReply} />}
       modalTitle="Markera som vidarebefordrad?"
-      onConfirm={() => dispatch(forwardCertificate(true))}
+      onConfirm={() => dispatch(forwardCertificate({ certificateId: certificateId, version: version, forward: true }))}
```
SamuelTilly, line 58: Looks like `forwardCertificate` needs to updated to accept this new interface.

image

mhornfeldt: You're right! I had forgotten to commit two files. Now they're up :)

SamuelTilly:
When `handleForwardCertificate` is called from `ListPage` it seems to be missing `version` parameter and the request fails with `UNKNOWN_INTERNAL_PROBLEM`.

Everything looks to be working as expected from `CertificatePage` 👍

mhornfeldt:
> When `handleForwardCertificate` is called from `ListPage` it seems to be missing `version` parameter and the request fails with `UNKNOWN_INTERNAL_PROBLEM`.
> 
> Everything looks to be working as expected from `CertificatePage` 👍

I've tested with the ListPage and everything is working correctly. If you want to test the code locally then you need the changes from backend also.

mww-sopra-steria (approved):

PR #409, opened 2022-07-08
INTYGFV-14626: Add scroll to search and list page

SamuelTilly:


SamuelTilly:
> Could it be refactored to be used in the CertificatePage?

Hmm, probably. I'll give it a try :)

PR #411, opened 2022-07-12
INTYGFV-14638: Header link to portal site for private practitioners

SamuelTilly:


mww-sopra-steria (changes requested):
The WebcertHeaderUser could really use a couple of tests to verify the additional behaviours that you have added. For example: suggest that you add tests to verify that the link is displayed when it should and that it opens it on click.

mww-sopra-steria (approved):

PR #418, opened 2022-07-21
INTYGFV-14624: Modals should render outside parent DOM node

SamuelTilly:


packages/common/src/components/utils/Modal/ModalBase.tsx
```diff
@@ -28,15 +24,29 @@ interface Props {
...
 
-  return (
-    <>
-      <FocusTrap active={open}>
-        <div tabIndex={0}>
-          <Backdrop className="ic-backdrop iu-lh-body" onClick={handleClose} />
+  const rootElement = document.getElementById('modalRoot')
+
+  if (!rootElement) {
+    return <></>
```
perfli, line 36: Shouldn't we return null instead?

perfli (approved):
Since this affects all modals, have you tested different modals to make sure they still work as expected?

PR #427, opened 2022-07-28
Fix linting issues

SamuelTilly:


packages/common/src/types/certificate.ts
```diff
@@ -300,19 +300,19 @@ export interface TextValidation extends CertificateDataValidation {
   // expression: string // '$6.1 > 5000' // Kan allt beskrivas med expression?
 }
 
-export interface ShowValidation extends CertificateDataValidation {}
+export type ShowValidation = CertificateDataValidation
```
perfli, line 303: I don't have a very good understanding of the differences between types and interfaces in TS but do we really want to make the change to type here?

SamuelTilly: Essentially it would have the same effect since we are extending with an empty interface.

This is fixed automatically by the linter from using this rule https://github.com/typescript-eslint/typescript-eslint/blob/v4.27.0/packages/eslint-plugin/docs/rules/no-empty-interface.md

perfli (approved):

PR #432, opened 2022-07-29
Reviewdog should not fail on warnings

SamuelTilly:


SamuelTilly:
Without "fail_on_error" the eslint part of github actions won't fail even if there is an error, but at least it wont fail when there is only warnings 🤷‍♂️

perfli (approved):

PR #444, opened 2022-08-17
INTYGFV-14457: Added logic for normal origin warning

mhornfeldt:


packages/webcert/src/feature/certificate/Modals/WarningNormalOriginModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const user = useSelector(getUser)
+  const resourceLink = useSelector(getUserResourceLink(ResourceLinkType.WARNING_NORMAL_ORIGIN))
+  const [open, setOpen] = useState(true)
+
+  const handleClose = () => {
+    if (!user) {
+      return
+    }
+    setOpen(false)
+    window.sessionStorage[user.loggedInUnit.unitId] = true
```
HelenaEdblad, line 16: I'm not so used to using sessionStorage so I'm just curious why we are using it here?

packages/webcert/src/store/user/userSelectors.ts
```diff
@@ -15,6 +15,9 @@ export const selectIsLoadingUser = (state: RootState): boolean => state.ui.uiUse
 
 export const getUserResourceLinks = (state: RootState): ResourceLink[] => state.ui.uiUser.links
 
+export const getUserResourceLink = (type: ResourceLinkType) => (state: RootState): ResourceLink | undefined =>
```
HelenaEdblad, line 18: Nice!

packages/webcert/src/feature/certificate/Modals/WarningNormalOriginModal.tsx
mhornfeldt: It was used in the original client to make sure that the modal only opens once per session.

HelenaEdblad (approved):

PR #447, opened 2022-08-17
INTYGFV-14454: Add info text about chosen unit in care provider modal

HelenaEdblad:


packages/common/src/components/Table/ExpandableTableRow.tsx
```diff
@@ -1,13 +1,19 @@
 import ArrowToggle from '@frontend/common/src/components/utils/ArrowToggle'
 import React, { Fragment, useState } from 'react'
+import styled from 'styled-components'
+
+const StyledButton = styled.button<{ disabled: boolean }>`
+  cursor: ${(props) => (props.disabled ? 'default' : 'pointer')};
```
mhornfeldt, line 6: Is this necessary when the button has the attribute disabled, shouldn't it come automatically?

mhornfeldt (changes requested):

packages/common/src/components/Table/ExpandableTableRow.tsx
HelenaEdblad: It didn't come automatically, that's why I added that css.

mhornfeldt: The default cursor for a disabled button is a pointer so I don't think you need to add logic for which cursor should be applied. If you have added som classnames which overrides the cursor: default styling then that should be the only thing you need to add.

HelenaEdblad: Ah, I removed the logic and added some classes instead now.

SamuelTilly (changes requested):

packages/webcert/src/feature/careProvider/CareProviderModalContent.tsx
```diff
@@ -63,27 +73,43 @@ export const CareProviderModalContent: React.FC = () => {
     }
   }
 
+  const getUnitName = (unit: Unit) => {
+    const chosenUnitInfoText = isLoggedInUnit(unit) && ' (vald enhet)'
+
+    if (isLoggedInUnit(unit)) {
+      return unit.unitName + chosenUnitInfoText
+    }
```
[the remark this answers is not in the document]
HelenaEdblad: Oh, of course. Fixed!

SamuelTilly (approved):

PR #449, opened 2022-08-18
INTYGFV-14731: Style fixes in header

HelenaEdblad:


packages/webcert/src/components/header/WebcertHeaderUser.tsx
```diff
@@ -40,15 +40,8 @@ const WebcertHeaderUser: React.FC = () => {
...
-          <UserRole>
-            {user.role}
-            {user.protectedPerson && (
-              <>
-                {' - '} <ProtectedPersonDoctorModal />
-              </>
-            )}
-          </UserRole>
+          <span>{`${user.name} - ${user.role}`}</span>
+          <UserRole>{user.protectedPerson && <ProtectedPersonDoctorModal />}</UserRole>
```
mhornfeldt, line 44: Why is the wrapper called UserRole if it only includes the protected person modal? Also, isn't this gonna result in an empty tag if the user doesn't have the protected person flag?

HelenaEdblad: Good question, I don't know why it's called UserRole but I changed it to StyledSpan and moved the condition outside of the  span.

SamuelTilly (approved):

PR #451, opened 2022-08-19
INTYGFV-14455: Change modal title

HelenaEdblad:


packages/webcert/src/feature/subscription/SubscriptionWarningModal.tsx
```diff
@@ -60,7 +60,7 @@ const SubscriptionWarningModal: React.FC = () => {
       handleClose={() => {
         return
       }}
-      title="Automatisk avtalskontroll sker för samtliga användare"
```
mhornfeldt, line 63: This text should come from the resource link

mhornfeldt (changes requested):

packages/webcert/src/feature/subscription/SubscriptionWarningModal.tsx
HelenaEdblad: Since the title and text in the modal doesn't have any business logic we decided that it was going to stay in the frontend code to make it easier to maintain.

mhornfeldt (approved):

PR #453, opened 2022-08-19
INTYGFV-14720: List search should not be performed if list type of filter does not match type of config

mhornfeldt:


SamuelTilly (approved):
Code looks good and bug is gone 👍

PR #458, opened 2022-08-23
INTYGFV-14453: Fix care provider modal

HelenaEdblad:


packages/webcert/src/feature/careProvider/CareProviderModal.tsx
```diff
@@ -37,7 +37,7 @@ const CareProviderModal: React.FC = () => {
   })
 
   const handleClose = () => {
-    dispatch(updateIsCareProviderModalOpen(false))
+    !chooseUnitLink && dispatch(updateIsCareProviderModalOpen(false))
```
mhornfeldt, line 40: Test should be added for this

HelenaEdblad: Please check if the test I wrote looks ok?

packages/webcert/src/feature/careProvider/CareProviderModal.test.tsx
```diff
@@ -140,5 +140,13 @@ describe('Care provider modal', () => {
       renderComponent()
       expect(screen.getByText('Välj vårdenhet')).toBeInTheDocument()
     })
+
+    it('should not allow closing the modal if choose unit resource link exists', () => {
+      testStore.dispatch(updateUserResourceLinks(getChooseUnitResourceLink()))
+
+      renderComponent()
+      userEvent.click(document.body)
+      expect(testStore.getState().ui.uiUser.isCareProviderModalOpen).toEqual(true)
```
mhornfeldt, line 149: This test is closely linked to the implementation of the component. If the isCareProviderModalOpen would be refactored and removed then the test would not make sense. Why don't you test that the modal is open by checking if the modal title exists for example?

mhornfeldt (approved):

PR #462, opened 2022-08-24
INTYGFV-14750: Fix issue where lists are updating after the active type has changed

SamuelTilly:


SamuelTilly:
After some testing, it's clear this needs a bit more work

SamuelTilly:
@HelenaEdblad , @mhornfeldt This is ready for CR now.

mhornfeldt (approved):

PR #475, opened 2022-08-30
INTYGFV-14670: Private practitioner should not show care provider name

mhornfeldt:


packages/webcert/src/components/header/WebcertHeaderUnit.tsx
```diff
@@ -40,7 +46,7 @@ const WebcertHeaderUnit: React.FC = () => {
     return (
       <Wrapper>
         <span>
-          {user.loggedInCareProvider.unitName} - {user.loggedInUnit.unitName}
+          {!privatePractitioner && `${user.loggedInCareProvider.unitName} - `} {user.loggedInUnit.unitName}
```
HelenaEdblad, line 49: I'm not sure if I understand the Jira. Should the care provider name really be taken away and only show the unit name?

mhornfeldt: https://inera.atlassian.net/browse/INTYGFV-14670

mhornfeldt: I think the reasoning behind removing the care provider name is that a private practitioner does not belong to a care provider and therefore will have the unit name as care provider name.

HelenaEdblad (approved):

PR #477, opened 2022-08-31
INTYGFV-14594: Redirect user to error page when certificate is not found

HelenaEdblad:


packages/webcert/src/store/certificate/certificateMiddleware.ts
```diff
@@ -155,7 +155,7 @@ const handleGetCertificateSuccess: Middleware<Dispatch> = ({ dispatch }) => () =
 }
 
 const handleGetCertificateError: Middleware<Dispatch> = ({ dispatch }) => () => (action: AnyAction): void => {
-  if (action.payload.error.errorCode === ErrorCode.UNKNOWN_INTERNAL_PROBLEM.toString()) {
```
mhornfeldt, line 158: Shouldn't you add the new error code instead of deleting the old one? I don't think this solution will present the correct error message when the network request is blocked, which is what the jira originally was aiming to fix.

mhornfeldt (changes requested):

packages/webcert/src/store/certificate/certificateMiddleware.ts
HelenaEdblad: Aha, ok, is the new solution better?

mhornfeldt (approved):

PR #479, opened 2022-09-01
INTYGFV-14753: Change title and message to concurrent modification error

HelenaEdblad:
Does it makes sence changing this component or is the solution not that easy?


HelenaEdblad:
Realized this is not the correct solution and I will need more testing and work on this.

HelenaEdblad:
Closed pull request because I realized it required backend changes instead of frontend.

PR #481, opened 2022-09-02
INTYGFV-14734: Display correct error message for list page

SamuelTilly:


packages/webcert/src/page/ListPage.tsx
```diff
@@ -105,10 +106,17 @@ const ListPage: React.FC<Props> = ({ type, excludePageSpecificElements }) => {
       return noDraftsImage
     }
   }
+  console.log(listError)
```
HelenaEdblad, line 109: Remove console.log :)

packages/webcert/src/page/ListPage.tsx
```diff
@@ -105,10 +106,17 @@ const ListPage: React.FC<Props> = ({ type, excludePageSpecificElements }) => {
...
 
   const getList = () => {
-    if (error) {
-      return <InfoBox type="error">Sökningen kunde inte utföras.</InfoBox>
+    if (listError) {
+      return (
+        <InfoBox type="error">
+          <div>
```
HelenaEdblad, line 115: Is this div necessary?

HelenaEdblad (changes requested):

packages/webcert/src/page/ListPage.tsx
SamuelTilly: I could look into moving it into the InfoBox component, would probably be better there. Without this div, we get an undesired layout since InfoBox only expects one child element.

HelenaEdblad: Aha, could <></> work? :)

SamuelTilly: Nope, that will still create multiple children for the HTML element, so the CSS would still have the same problem.

HelenaEdblad: Ok then :)

HelenaEdblad (approved):

PR #484, opened 2022-09-14
INTYGFV-14467: Add support for signing certificates with bankid

SamuelTilly:
For mobile texts to appear, minor changes to backend are required:
https://github.com/sklintyg/common/pull/589
https://github.com/sklintyg/webcert/pull/767


packages/common/src/types/certificate.ts
```diff
@@ -420,3 +420,12 @@ export interface IcfTitles {
     common: string[]
   }
 }
+
+export enum CertificateSignStatus {
+  INITIAL = '',
+  UNKNOWN = 'OKAND',
+  PROCESSING = 'BEARBETAR',
```
IamFriberg, line 427: Hur kommer det sig att vi blandar svenska och engelska enumar här?

packages/webcert/src/store/certificate/certificateMiddleware.ts
```diff
@@ -317,25 +322,94 @@ const handleStartSignCertificate: Middleware<Dispatch> = ({ dispatch, getState }
 
   const signingMethod = getState().ui.uiUser.user.signingMethod
 
-  if (signingMethod === SigningMethod.FAKE) {
-    dispatch(fakeSignCertificate)
-  } else if (signingMethod === SigningMethod.DSS) {
+  switch (signingMethod) {
```
IamFriberg, line 325: Do we have any unittests for this? :)

packages/webcert/src/feature/certificate/Modals/SignCertificateModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+  return (
+    <p>
+      Intyget har inte signerats. Detta beror antingen på ett tekniskt fel eller att signeringen avbrutits. <br />
+      Prova igen om en stund. Om problemet kvarstår, kontakta i första hand din lokala IT-avdelning och i andra hand{' '}
+      <WCDynamicLink linkKey="ineraKundserviceAnmalFel" />.
+    </p>
+  )
+}
+
+export const SignCertificateModal: React.FC = () => {
```
HelenaEdblad, line 177: I think we need tests for this modal :)

HelenaEdblad (changes requested):

packages/common/src/types/certificate.ts
SamuelTilly: Good question. I believe we need to have this translation, currently this code is sharing its backend endpoint with the old client, so the values from the backend are in Swedish while the codebase for this project is English. Could be that the translation is perhaps better suited as a backend task.

Should/could we write a translation endpoint that gives us English values? @mww-sopra-steria

packages/webcert/src/components/error/modals/errorUtils.ts
```diff
@@ -22,7 +22,7 @@ export const uuidv4 = (): string => {
   })
 }
 
-export const messageSubstring = (activeError: ErrorData): string | undefined => {
+export const messageSubstring = (activeError: { message?: string }): string | undefined => {
```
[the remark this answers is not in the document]
SamuelTilly: This won't introduce a regression bug. Instead of only allowing for the specific `ErrorData` we allow for a more generic object containing message that will still function with the `ErrorData` type :)

packages/webcert/src/store/certificate/certificateMiddleware.ts
SamuelTilly: Don't think there are any currently, I'll look into adding that.

packages/webcert/src/feature/certificate/Modals/SignCertificateModal.tsx
SamuelTilly: Good call, I'll add a couple of tests for the modal

packages/webcert/src/components/error/modals/errorUtils.ts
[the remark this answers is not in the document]
HelenaEdblad: Ok, nice :)

packages/common/src/types/certificate.ts
mww-sopra-steria: The enums are part of legacy code and I wouldn't change it as it would require changes in both backend and current angularjs client.

packages/webcert/src/store/certificate/certificateMiddleware.ts
SamuelTilly: unit tests have been added.

packages/webcert/src/feature/certificate/Modals/SignCertificateModal.tsx
SamuelTilly: unit tests have been added.

HelenaEdblad (approved):

packages/common/src/types/certificate.ts
SamuelTilly: Alright, I would like to keep the value translated in this enum then, since I believe it would be much worse if we have Swedish words randomly showing up in other contexts where `CertificateSignStatus` is used. Keeping the Swedish to a minimum :)

PR #488, opened 2022-09-16
INTYGFV-14127  new RegEx for date fasibility check in dateUtil

SopraErik:
Changed the RegEx for the date feasibility check in dateUtil to accept only dates between 1900-01-01 and 2099-12-12, same as in the backend.

Also added some tests to verify this


SamuelTilly:
Doesn't look like we get the desired results here.

Expected result:
image

Actual result:
image

SamuelTilly:
Almost there now, just expecting a different error message that informs about date being too far back or forward.

image

SamuelTilly:
Validation error appears for wrong field and 2099-12-12 should be invalid, only triggers on 2099-12-13 now.
image

packages/common/src/utils/dateUtils.ts
```diff
@@ -29,6 +33,15 @@ export const getValidDate = (dateString: string | undefined) => {
   }
 }
 
+export const getReasonableDate = (dateString: string | undefined) => {
+  if (!dateString) return
+  
+  let dateValue = getValidDate(dateString)
+  if (dateValue) {
+    if (!isBefore(dateValue!, _minAllowedDate) && !isAfter(dateValue!, _maxAllowedDate)) return dateValue
```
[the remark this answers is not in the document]
SamuelTilly: No need for non-null assertions here, please change `dateValue!` to `dateValue`.

packages/common/src/utils/dateUtils.ts
```diff
@@ -29,6 +33,16 @@ export const getValidDate = (dateString: string | undefined) => {
   }
 }
 
+export const getReasonableDate = (dateString: string | undefined) => {
+  if (!dateString) return
+  
+  let dateValue = getValidDate(dateString)
```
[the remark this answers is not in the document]
SamuelTilly: This should also be changed.

packages/common/src/utils/dateUtils.test.ts
```diff
@@ -63,17 +64,88 @@ const configList: ConfigUeCheckboxDateRange[] = [
...
-    const date = getValidDate(validDateStringDashes)
+    const date = getValidDateFormat(validDateStringDashes)
     expect(date).toBeTruthy()
   })
 
   it('get valid date without dashes string', () => {
     const validDateString = '20210408'
+    const date = getValidDateFormat(validDateString)
+    // console.log(date)
+    expect(date).toBeTruthy()
```
SamuelTilly, line 75: Would prefer to check on the actual return on this so that instead of "truthy" we check the exact output `expect(date).toEqual('20210408')` and instead of "falsy" we should check that it's undefined `expect(date).toBeUndefined()`

packages/common/src/utils/dateUtils.ts
```diff
@@ -4,6 +4,10 @@ import { ConfigUeCheckboxDateRange, replaceDecimalSeparator } from '..'
 
 export const _dateReg = /[1-2][0-9]{3}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])/
 export const _dateRegDashesOptional = /[1-2][0-9]{3}-?(0[1-9]|1[0-2])-?(0[1-9]|[1-2][0-9]|3[0-1])/
+
+export const _minAllowedDate = new Date(1900, 0, 1)
+export const _maxAllowedDate = new Date(2099, 11, 12)
```
SamuelTilly, line 9: Shouldn't this be `new Date(2099, 12, 12)`

packages/common/src/utils/dateUtils.test.ts
```diff
@@ -63,15 +64,86 @@ const configList: ConfigUeCheckboxDateRange[] = [
...
+  it('get reasonable 1900 date with valid dashes string', () => {
+    const validDateStringDashes = '1901-04-08'
+    const date = getValidDate(validDateStringDashes)
+    expect(date).toEqual(new Date(1901, 3, 8))
+  })
+
+  it('get reasonable 1900 date without dashes string', () => {
+    const validDateString = '19010408'
     const date = getValidDate(validDateString)
     // console.log(date)
```
SamuelTilly, line 93: Vi skulle kunna städa bort dom här, sen tycker jag det ser bra ut och redo för en merge 👍

SamuelTilly (approved):

PR #494, opened 2022-09-23
INTYGFV-14789: Create certificate component for message

HelenaEdblad:


SamuelTilly (changes requested):
Rename AlertMessage to UeMessage and move to "src\feature\certificate\Inputs" together with the other Ue components for consistency. Other than that, it looks great 👍

PR #495, opened 2022-09-23
INTYGFV-14792: Add support for certificate AUTO_FILL_VALIDATION

SamuelTilly:


packages/webcert/src/store/certificate/certificateReducer.ts
```diff
@@ -269,9 +271,13 @@ const certificateReducer = createReducer(getInitialState(), (builder) =>
...
 
-      state.certificate.data[action.payload].disabled = true
-      ;(state.certificate.data[action.payload].value as ValueCode).id = ''
-      ;(state.certificate.data[action.payload].value as ValueCode).code = ''
+      const question = state.certificate.data[action.payload]
+
+      question.disabled = true
+      if (question?.value?.code) {
+        ;(question.value as ValueCode).id = ''
+        ;(question.value as ValueCode).code = ''
```
HelenaEdblad, line 279: What does the ; mean? Can you elaborate?

HelenaEdblad (changes requested):

mww-sopra-steria (dismissed):
We have now done an implementation in backend. Used this PR as a guide. 

While implementing we made a change in the AUTO_FILL_VALIDATION "model". Instead of having a AutoFillValidationType, we used the CertificateValueType and believe that the AutoFillValidationType could be redundant.

What do you think about this? See screenshot of json-data retrieved from backend.

image

packages/webcert/src/store/certificate/certificateReducer.ts
SamuelTilly: Good question, for some reason prettier want's this, I guess it's having issues with parsing instances where type assignment and variable assignment are done simultaneous. I am not entirely sure.

2022-10-07 10_56_45-● certificateReducer ts - frontend - Visual Studio Code

SamuelTilly:
> We have now done an implementation in backend. Used this PR as a guide.
> 
> While implementing we made a change in the AUTO_FILL_VALIDATION "model". Instead of having a AutoFillValidationType, we used the CertificateValueType and believe that the AutoFillValidationType could be redundant.
> 
> What do you think about this? See screenshot of json-data retrieved from backend.
> 
> image

This makes sense! I will replace `AutoFillValidationType` with `CertificateDataValueType` for this PR as well.

HelenaEdblad (approved):

PR #499, opened 2022-09-27
INTYGFV-14816: Add date picker component for certificates

HelenaEdblad:


SamuelTilly (changes requested):
I believe we should change the name for the component to `UeDate`, to be consistent with `UeCheckboxDate` etc. And also change the `UE_DATEPICKER` to `UE_DATE` for the same reason as defined in the jira.

SamuelTilly (approved):

PR #504, opened 2022-10-03
RTK Query Experiment

SamuelTilly:
When calling the API using the apiMiddleware it expects a callback function for updating "functionDisablers", this is used to disable every related functionality to the requests that is being done. It does this with a toggle function that on the first call generates an identifying object and stores that in an array at the start of a request, when the request is done it calls the same function and that object is removed. The problem is that if two or more API calls are made at the same time, it will call the function while an identifying object is still being generated, which results in a list that only contains the last generated object.

2022-09-29 13_53_54-Webcert

The first two calls in these screenshots are when the object is being generated, and the first object is lost because the second call updates the `state.functionDisablers` with an empty array:

https://github.com/sklintyg/frontend/blob/364c32bfbb18c3a7cdfa6d9596f5b3af7e7c8a77/packages/webcert/src/store/question/questionReducer.ts#L140-L142

It gets updated twice because it will (when conditions are right) be called every time the certificate gets updated, and that happens to be twice on initial page-load, `dispatch(getQuestions(action.payload.metadata.id))`  here: 
https://github.com/sklintyg/frontend/blob/364c32bfbb18c3a7cdfa6d9596f5b3af7e7c8a77/packages/webcert/src/store/question/questionMiddleware.ts#L154-L171

## Enter "RTK Query"

https://redux-toolkit.js.org/rtk-query/overview

While this is a big refactoring, if we start using RTK Query (already installed in this project due to being part of redux-toolkit) we not only solve this issue, we can also remove a majority of the boilerplate code when dealing with API requests for Redux. In this example I only replace the code that deals with questions, answers and complements without having to touch certificate logic, but going forward we could start writing code using RTK Query and significantly reduce the complexity of this project.

RTK Query introduces:
 
* Generated hooks.
* Built in request caching.
* Loading, success and error states.
* Much less code.


mww-sopra-steria:
Out of curiosity I looked a little bit on this PR. As you say it includes a pretty big refactoring, but it also changes a central part of the client architecture. I do see the flaw in "functionDisablers" and the challenge to deal with multiple api-calls. But the impact of this architectural change should not be underestimated. Could we reason together around this just to make sure that we are moving in the right direction here and consider any potential consequence?

SamuelTilly:
> Out of curiosity I looked a little bit on this PR. As you say it includes a pretty big refactoring, but it also changes a central part of the client architecture. I do see the flaw in "functionDisablers" and the challenge to deal with multiple api-calls. But the impact of this architectural change should not be underestimated. Could we reason together around this just to make sure that we are moving in the right direction here and consider any potential consequence?

I agree, this is certainly a bit much for a bugfix. I haven't even started to look at how this will affect the existing tests, but I suspect there could be a lot of work. Since modern redux is recommended to be used this way, I wanted to investigate what that would mean for our codebase. I have and will continue to investigate how to fix this bug without having to change everything over to RTKQ, but the problem with our current solution is that it's very hard to know how things will be affected when events are triggered, and it's a lot of events that get triggered.

PR #512, opened 2022-10-14
wrapping infobox in styled div with negativ top margin to adjust for …

SopraErik:
going with the somewhst ugly solution of wrapping infobox in styled div with negativ top margin to adjust for too large spacing between question and message


SamuelTilly:
This works, but perhaps we could come up with a bit more general solution that fixes a couple of more places that are having issues.
 
2022-10-14 10_45_57-Webcert _ Certificate - Dödsbevis ⋅ Storybook
2022-10-14 10_46_13-Webcert _ Certificate - Dödsbevis ⋅ Storybook

And if the message component will ever be at the top of a category, it will create a bit of broken layout.

2022-10-14 10_49_26-Webcert _ Certificate - Dödsbevis ⋅ Storybook

SamuelTilly (approved):

PR #514, opened 2022-10-17
INTYGFV-14847: Create component for showing patient address

HelenaEdblad:


packages/webcert/src/feature/certificate/Certificate.tsx
```diff
@@ -79,6 +83,7 @@ const Certificate: React.FC = () => {
           </InfoBox>
         )}
         <ResponsibleHospName />
+        {showPatientAddress && <PatientAddress />}
```
SamuelTilly, line 86: Needs to have `CategoryWrapper` for proper padding between categories.

```tsx
<CategoryWrapper>
  <PatientAddress />
</CategoryWrapper>
```

packages/webcert/src/feature/certificate/PatientAddress/PatientAddress.tsx
```diff
@@ -0,0 +1,86 @@
...
+import CategoryHeader from '../Category/CategoryHeader'
+import CategoryTitle from '../Category/CategoryTitle'
+import QuestionWrapper from '../Question/QuestionWrapper'
+
+const Wrapper = styled.div`
+  align-items: center;
+`
+
+const ZipCodeInput = styled(TextInput)`
+  max-width: 6.5em;
```
SamuelTilly, line 15: Could be a tiny bit larger to allow all characters to show on different systems

image

SamuelTilly (changes requested):

packages/webcert/src/feature/certificate/PatientAddress/PatientAddress.tsx
HelenaEdblad: It has the same measurements as the field for zip code in the component for care unit address. The field will probably only hold Swedish zip codes so I will leave it this size for now. :)

SamuelTilly (approved):

PR #523, opened 2022-10-25
Feature/intygfv 14849 - modals when signing certificate

SopraErik:
NB! This PR also handles INTYGFV-14851

Support for modals when signing certificate
Modals are controlled by resource links
if resource link SIGN_CERTIFICATE_CONFIRMATION is present, the sign button will be visible and  a modal will always be displayed when the button is clicked
If resource link SIGN_CERTIFICATE **is not** present, a confirmation button will not be displayed, and the user may only cancel 
(eg. when a signed death certificate already exists)
If resource link SIGN_CERTIFICATE **is** present the user may choose to continue with signing or abort
(eg. when a deat certificate draft exists)
If no SIGN_CERTIFICATE_CONFIRMATION  exists, but SIGN_CERTIFICATE exists, the sign buttn will show, but no modal will be displayed


packages/webcert/src/feature/certificate/Buttons/SignAndSendButton.tsx
```diff
@@ -1,34 +1,65 @@
...
+        confirmButtonText={name}
+        onConfirm={handleConfirm}
+        disabled={isValidating || isSigning || !enabled || functionDisabled}
+        hideConfirmButton={!canSign}>
+        {body ? (
+          <div>
+            <p>{body}</p>
+          </div>
+        ) : (
+          <></>
```
HelenaEdblad, line 46: I think you can write it like this to avoid the empty element:

```
body && (
  <div>
    <p>{body}</p>
  </div>
)
```

HelenaEdblad (changes requested):

SamuelTilly (approved):
Some minor tweaks could be made, but looks really nice! 🥇

SopraErik:
@HelenaEdblad can you please review my changes so that I can merge the branch

HelenaEdblad:
Since Samuel approved the PR I will dismiss my review.

HelenaEdblad (approved):

PR #524, opened 2022-10-26
modified onChange so correct values are dispatched

SopraErik:
Modified the onchange events in uncertain date so that the correct values are dispatched


ChrilleSigerhed (approved):
Well done, now UNCERTAIN_DATE validates correct in frontend and backend!

PR #531, opened 2022-10-31
INTYGFV-14921:  Added function missingValue to parseExpression to handle undefined/null boolean values

ChrilleSigerhed:


packages/common/src/utils/validationUtils.ts
```diff
@@ -57,6 +57,10 @@ export const parseExpression = (
       }
       case CertificateDataValueType.BOOLEAN: {
         const valueBoolean = element.value as ValueBoolean
+
+        if ((valueBoolean.selected === undefined || valueBoolean.selected === null) && adjustedExpression.includes('not')) {
```
IamFriberg, line 61: How come we check "not"?

IamFriberg (changes requested):

packages/common/src/utils/validationUtils.ts
ChrilleSigerhed: Given we have two expressions, one that checks if a statement is true one that checks if its false, eg ($dodsdatumSakert & !$dodsdatumSakert).

This gets translated to "dodsdatumSakert" and "not dodsdatumSakert" in getExpression(). 

The first expression gets triggered if the statement is true, and the second if its false. Since the second statement (!$dodsdatumSakert) gets triggered if the statement is false, I added logic that returns true to not trigger negative expressions that are null/undefined.

mww-sopra-steria (approved):
Nice!

IamFriberg (approved):

PR #532, opened 2022-11-01
bug/INTYGFV-14923 - updated typeahead to dispatch only if value is changed. Also introduc…

SopraErik:
updated typeahead to dispatch only if value is changed. 
Also introduced new config property for UeTypeahead for placeholder for better UX


SopraErik:
@mww-sopra-steria , @ChrilleSigerhed etc. I have updeted the config for UeTypeahead to incorporate a placeholder:
`export interface ConfigUeTypeahead extends CertificateDataConfig {
  id: string
  typeAhead: string[]
  text: string
  label: string
  placeholder?: string
}`
This to make the UX more intuitive

mww-sopra-steria (changes requested):
Did some tests but am getting the same behaviour in the client. Maybe we could do a test together and verify that we are dealing with the correct problem.

mww-sopra-steria (approved):
Nice! I will approve! 

A thought, is it possible to add some test on the changed behaviour of the component? If so, please do before merging.

PR #533, opened 2022-11-01
bug/INTYGFV-14925: Using property to check if question will be in read view

UlrikaSopra:


packages/common/src/components/UvText/UvText.tsx
```diff
@@ -265,7 +265,7 @@ const UvText: React.FC<Props> = ({ question }) => {
         if (textValueDate.date != null && textValueDate.date.length > 0) {
           displayText = textValueDate.date
         } else {
-          displayText = ''
+          displayText = 'Ej angivet'
```
ChrilleSigerhed, line 268: If uncertain date is selected for example, will this display "Ej angivet" for dodsdatum?

SopraErik: No, uncertain date is handled in the next case statement (lines 272-279). Maybe this needs to be updated too?

SopraErik: I dont think we need to set displayText if no value is present, since it is set to "Ej angivet" by default in line 188

SopraErik (changes requested):
I dont think we need to set displayText if no value is present, since it is set to "Ej angivet" by default in line 188

mww-sopra-steria (approved):
Nice! See though if there are possibility to add some tests before merging.

packages/common/src/utils/validationUtils.test.ts
```diff
@@ -1152,8 +1152,6 @@ describe('Set initial values to a certificate', () => {
 
       expect(certificate.data['1.1'].readOnly).toBe(true)
       expect(certificate.data['1.2'].readOnly).toBe(true)
-      expect(certificate.data['1.1'].visible).toBe(true)
```
ChrilleSigerhed, line 1155: Instead of removing the visibility check, would it be good to keep makeing that control and instead remove the visibility = false in function clearValues() that gets called in the start of each test? Or change it so it sets visible = undefined

packages/common/src/components/UvText/UvText.test.tsx
```diff
@@ -82,6 +82,13 @@ describe('UvText', () => {
     expect(screen.getByText(/Ej angivet/i)).toBeInTheDocument()
   })
 
+  it('displaying text value if question is visible', () => {
```
ChrilleSigerhed, line 85: Since the logic that was implemented intended to hide elements, should we implement a test that verifies that an element is not showed if visibel is set to false?

packages/common/src/utils/validationUtils.test.ts
UlrikaSopra: ok, i will change it to === undefined

ChrilleSigerhed (approved):
Great job! Looks good!

SopraErik (approved):

PR #536, opened 2022-11-01
INTYGFV-14911: Add text to modal for revoking death certificate

HelenaEdblad:


packages/webcert/src/feature/certificate/Buttons/Test/RevokeCertificateButton.test.tsx
```diff
@@ -180,4 +198,47 @@ describe('Revoke continue button', () => {
       type: '[CERTIFICATE] Revoke certificate',
     })
   })
+
+  describe('isDodsbevis', () => {
+    const mockDispatchFn = jest.fn()
```
mww-sopra-steria, line 203: Can use dispatchHelperMiddleware instead of using mocks. The dispatchHelperMiddleware just stores the different actions that has been dispatched and exposes them for you to assert/expect what you need. 

Here is just an example from another test:
```
    it('should throw hide action when hidden', async () => {
        const certificate = showValidation(false, true)
        testStore.dispatch(updateCertificate(certificate))

        testStore.dispatch(validateCertificateInFrontEnd(certificate.data['1.1']))

        const hideCertificateDataElementAction = dispatchedActions.find((action) => hideCertificateDataElement.match(action))
        expect(hideCertificateDataElementAction && hideCertificateDataElementAction.payload).toBe('1.2')
      })
```

packages/common/src/utils/test/certificateTestUtil.ts
```diff
@@ -504,6 +504,91 @@ export const getCertificateWithQuestion = (question: CertificateDataElement): Ce
   }
 }
 
+export const getDeathCertificate = ({ links = [] }: { links?: ResourceLink[] } = {}): Certificate => {
```
mww-sopra-steria, line 507: Suggest that you use the fakers to create your needed testdata. The certificateTesetUtil.ts should been seen as legacy and something we should move away from. So it could be good not to add testdata to it.

example how to do what you want in this scenario:

`fakeCertificate({ metadata: fakeCertificateMetaData({ type: 'db' }) })`

Just add what extra data you need... And just create it together with your tests.

packages/webcert/src/feature/certificate/Buttons/RevokeCertificateButton.tsx
```diff
@@ -54,7 +58,7 @@ const RevokeCertificateButton: React.FC<Props> = ({ name, description, enabled,
       modalTitle={metadata?.status === CertificateStatus.LOCKED ? 'Makulera låst utkast' : 'Makulera intyg'}
       onConfirm={handleDispatch}
       confirmButtonText="Makulera">
-      <RevokeCertificateModalContent onChange={handleRevokeForm} type={metadata?.type} />
+      {!isDodsbevis ? <RevokeCertificateModalContent onChange={handleRevokeForm} type={metadata?.type} /> : <RevokeDodsbevisModalContent />}
```
mww-sopra-steria, line 61: Easier to read the code if you remove the negating operand ('!') and just change order to the revoke modals.

mww-sopra-steria (changes requested):
Nice! Few minor suggestions that could simplify things.

mww-sopra-steria (approved):

PR #542, opened 2022-11-03
Add end-to-end test github action

SamuelTilly:
This makes it possible to start local development against devtest backend service, removing the necessity to have nginx, docker, java etc. for frontend work. And makes it possible for end-to-end tests to run with the latest frontend changes against the latest deployed backend changes, adding extra security before deploying new code.


SamuelTilly:
The remaining issue is caused by the race-condition bug documented in INTYGFV-14823

HelenaEdblad (approved):

PR #546, opened 2022-11-04
INTYGFV-14932: Added PatientAddress to display on signed certificate

UlrikaSopra:


packages/webcert/src/store/certificate/certificateReducer.ts
```diff
@@ -1,5 +1,4 @@
 import {
```
SopraErik, line 1: Is the deletion here relevant? Isn't this to do with another feature?

SopraErik (changes requested):
Looks good, but I think the deletion in certificateReducer is probably some mistake?

packages/webcert/src/store/certificate/certificateReducer.ts
UlrikaSopra: Someone had forgotten to remove it and if I didn't remove it, I got an error in the browser

SopraErik: Ah, saw now that it's also in the master branch, and is never used there, so probably from some unmerged branch?
I'll approve the PR

SopraErik (approved):

PR #548, opened 2022-11-08
INTYGFV-14923: Added extra filter for typeahead result

UlrikaSopra:
Fixed the filtering of the list, adjusted the scrolling and the behavior when updating the list.


ChrilleSigerhed (commented):
Well done, now the sorting works fine!

If possible, it would be nice to add a test that verifies this functionality.

This component still dispatches while the suggested municipalities are showed, wich result in an unwanted update for the component and interrupts scrolling in typeahead component: 

image

Should we create a separate Jira for this or correct it in this Jira?

The scrolling behaviour when stepping down the suggested list using the arrow-keys feels in my opinion werid. Should this be looked over and corrected?

packages/webcert/src/feature/certificate/Inputs/UeTypeahead.tsx
```diff
@@ -68,9 +68,14 @@ const UeTypeahead: React.FC<Props> = ({ question, disabled }) => {
...
       }
-      setSuggestions(
-        questionConfig.typeAhead.filter((suggestion: string) => suggestion.toLowerCase().indexOf(newText.toLowerCase()) >= 0).sort()
-      )
+
+      const sortResult = [
+        ...questionConfig.typeAhead.filter((suggestion) => suggestion.toLowerCase().startsWith(newText.toLowerCase())),
+        ...questionConfig.typeAhead.filter((suggestion) => suggestion.toLowerCase().indexOf(newText.toLowerCase()) >= 0).sort(),
+      ]
+      const newResult = new Set(sortResult)
```
SamuelTilly, line 76: Instead of using Set and double filters to create a list of unique values, we could perhaps do one run of filtering.

```typescript
questionConfig.typeAhead.filter((suggestion) => {
  const haystack = suggestion.toLowerCase()
  const needle = newText.toLowerCase() 
  return haystack.startsWith(needle)  && haystack.indexOf(needle) >= 0
}))
```

SamuelTilly (changes requested):

packages/webcert/src/feature/certificate/Inputs/UeTypeahead.tsx
UlrikaSopra: It is necessary when you, for example, search for 'Ors', then you expect to get 'Orsa' at the top and also other words that contain 'Ors' . If we only use IndexOf we get this result (also added a sort()):
image

SamuelTilly (approved):

packages/webcert/src/feature/certificate/Inputs/UeTypeahead.tsx
UlrikaSopra: good idéa!

ChrilleSigerhed (changes requested):
Component is crashing when the value does not exist in the list it seems like. To recreate, try typing "hert".

image

ChrilleSigerhed (approved):
Awsome, the application no longer crashes and the scrolling behaviour is improved!

There is still a problem related to saving. It seems as if its required to select an option from the dropdown list in the component for the saving to kick in. To recreate, try typing in SopraSteria, and then refresh the page!

Since im going on vacation today I will approve, so that I dont block this PR. Even tho some changes are still needed for this component.

PR #558, opened 2022-11-15
INTYGFV-14877: added text when canceling a DOI certificate

melindalindgrensopra:


packages/webcert/src/feature/certificate/Buttons/RevokeDBandDoiModalContent.tsx
```diff
@@ -0,0 +1,56 @@
+import { InfoBox } from '@frontend/common'
+import _ from 'lodash'
+import React from 'react'
+import { useSelector } from 'react-redux'
+import { getCertificateMetaData, getIsLocked } from '../../../store/certificate/certificateSelectors'
+
+export const RevokeDBandDoiModalContent: React.FC = () => {
```
HelenaEdblad, line 7: Maybe doi also should be in capital letters, so it looks the same as DB? And 'and' should also start with a capital letter.
So change it to RevokeDBAndDOIModalContent

HelenaEdblad (changes requested):
Nice! Just one tiny change :)

HelenaEdblad (approved):

PR #576, opened 2022-11-29
INTYGFV-15013: added accordion functionality to controls

SopraErik:


packages/common/src/components/utils/Accordion.tsx
```diff
@@ -73,12 +96,14 @@ interface Props {
 const Accordion: React.FC<Props> = ({
   icon,
   title,
+  titleClose,
   titleId,
   description,
   additionalStyles,
   displayMandatory,
```
SamuelTilly, line 103: We can make sure this has a default value of false, so we don't need to assign `displayMandatory as boolean` in places.

packages/common/src/components/utils/Accordion.tsx
```diff
@@ -87,29 +112,43 @@ const Accordion: React.FC<Props> = ({
...
+          <h5 className={`iu-fs-100 iu-fw-bold iu-lh-body close ${additionalStyles}`}>{title}</h5>
+          <h5 className={`iu-fs-100 iu-fw-bold iu-lh-body open ${additionalStyles}`}>{titleClose}</h5>
         </StyledSummary>
       )
+    } else {
+      if (hasHeader) {
+        return (
+          <StyledSummary tabIndex={0} className="ic-expandable-button ic-inner ic-expandable-button--chevron iu-fs-400">
+            <Icon iconType={icon ? icon : ''} size={iconSize} />
+            <MandatoryIcon display={displayMandatory as boolean} additionalStyles={mandatoryIconAdditionalStyles} />
```
SamuelTilly, line 129: We should not assign `displayMandatory` to a boolean here, let's make sure it actually is a boolean instead.

packages/common/src/components/utils/Accordion.tsx
```diff
@@ -87,29 +112,43 @@ const Accordion: React.FC<Props> = ({
...
+        return (
+          <StyledSummary tabIndex={0} className="ic-expandable-button ic-inner ic-expandable-button--chevron iu-fs-400">
+            <Icon iconType={icon ? icon : ''} includeTooltip={includeIconTooltip} size={iconSize} />
+            <MandatoryIcon additionalStyles={mandatoryIconAdditionalStyles} display={displayMandatory as boolean} />
+            <h4 className={`${isCategory ? 'iu-fs-400' : 'iu-fs-300'} ${additionalStyles}`}>{title}</h4>
+          </StyledSummary>
+        )
+      }
     }
   }
```
SamuelTilly, line 143: We have a lot of repetition in this function, instead of having large if-statements we can toggle individual parts. For example `StyledSummary` always looks the same.

Something like:
```typescript
const getTitle = () => {
  if (isControl) {
    return <h5 className={`iu-fs-100 iu-fw-bold iu-lh-body close ${additionalStyles}`}>{title}</h5>
  } else if (hasHeader) {
    return <h4 className={`${isCategory ? 'iu-fs-400' : 'iu-fs-300'} ${additionalStyles}`}>{title}</h4>
  }
  return <h5 className={`iu-fs-200 iu-lh-body ${additionalStyles}`}>{title}</h5>
}

const getHeader = () => {
  return (
    <StyledSummary tabIndex={0} className="ic-expandable-button ic-inner ic-expandable-button--chevron iu-fs-400">
      <Icon iconType={icon ? icon : ''} includeTooltip={!hasHeader && includeIconTooltip} size={iconSize} />
      {displayMandatory && <MandatoryIcon additionalStyles={mandatoryIconAdditionalStyles} display={true} />}
      {getTitle()}
      {titleClose && <h5 className={`iu-fs-100 iu-fw-bold iu-lh-body open ${additionalStyles}`}>{titleClose}</h5>}
    </StyledSummary>
  )
}
```

This is just a quick example, could probably be refined even more?

packages/common/src/components/utils/Accordion.tsx
```diff
@@ -87,29 +112,43 @@ const Accordion: React.FC<Props> = ({
...
+            <MandatoryIcon display={displayMandatory as boolean} additionalStyles={mandatoryIconAdditionalStyles} />
+            <h5 className={`iu-fs-200 iu-lh-body ${additionalStyles}`}>{title}</h5>
+          </StyledSummary>
+        )
+      } else {
+        return (
+          <StyledSummary tabIndex={0} className="ic-expandable-button ic-inner ic-expandable-button--chevron iu-fs-400">
+            <Icon iconType={icon ? icon : ''} includeTooltip={includeIconTooltip} size={iconSize} />
+            <MandatoryIcon additionalStyles={mandatoryIconAdditionalStyles} display={displayMandatory as boolean} />
+            <h4 className={`${isCategory ? 'iu-fs-400' : 'iu-fs-300'} ${additionalStyles}`}>{title}</h4>
```
SamuelTilly, line 138: I wonder if `additionalStyles` is ever used for this component, seems strange that styling passed to this component can be placed on every header element.

packages/common/src/components/utils/Accordion.tsx
```diff
@@ -51,19 +51,42 @@ const StyledSummary = styled.summary`
   }
 `
 
-const StyledDetails = styled.details`
+interface DetailsProps {
+  title?: string
+  titleClose?: string
+}
+
+const StyledDetails = styled.details<DetailsProps>`
```
SamuelTilly, line 59: Looks like properties inside `DetailsProps` are never used here

packages/webcert/src/feature/certificate/Question/Question.tsx
```diff
@@ -89,48 +89,82 @@ const Question: React.FC<QuestionProps> = ({ id, className }) => {
...
+          icon={question.config.icon}
+          includeIconTooltip
+          titleId={question.id}
+          title={question.config.accordion.openText}
+          titleClose={question.config.accordion.closeText}
+          header={question.config.accordion.header}
+          isControl={true}>
+          {component}
+        </Accordion>
+      )
```
SamuelTilly, line 167: Instead of assigning to `component` we can just wrap this function.

```typescript
if (question.config.accordion) {
  return <Accordion ...>getUnifiedEditComponent(question, disabled)</Accordion>
}
return getUnifiedEditComponent(question, disabled)
```

packages/common/src/utils/faker/fakeCertificateData.ts
```diff
@@ -280,6 +281,27 @@ export const fakeTextAreaElement = (
...
+          openText: 'Visa fritextfält',
+          closeText: 'Dölj fritextfält',
+          header: faker.lorem.sentence(5),
+        },
+        ...data?.config,
+      },
+      value: { type: CertificateDataValueType.TEXT, text: 'Text', limit: 50, ...data?.value },
+    },
+    children
+  )
```
SamuelTilly, line 303: We should not create fakeElement function for individual configuration parameter. This should be removed and used with fakeTextAreaElement instead. 

```typescript
fakeTextAreaElement({
  config: {
    accordion: {  
      openText: 'Visa fritextfält',
      closeText: 'Dölj fritextfält',
      header: faker.lorem.sentence(5),
    }
  }
})
```

SamuelTilly (changes requested):

packages/common/src/components/utils/Accordion.tsx
SamuelTilly: Found a place that uses `additionalStyles`. Let's keep it where it already exists, but let's not add it to any new place, such as `closeText` header.

SamuelTilly (commented):
I think Accordion needs to become a much more generic component. Logic that is specific for "control", "category", "mandatory" etc. should not be part of this component. This component has suffered from a bit of feature creep, but is luckily currently only used in two places. Category and UeSickleavePeriod. We should make sure to separate Accordion from those and create specific components such as SickleaveAccordion, CategoryAccordion and QuestionAccordion that extends on Accordion and remove isCategory, isControl, Icon etc...

Accordion in its simplest form could be defined with only styled-component:

```typescript
import styled from 'styled-components'

const Accordion = styled.details.attrs({
  className: 'ic-expandable',
})``

export default Accordion
```

And AccordionHeader:
```typescript
import styled from 'styled-components'

const AccordionHeader = styled.summary.attrs({
  tabIndex: 0,
  className: 'ic-expandable-button iu-focus',
})``

export default AccordionHeader
```

SamuelTilly (approved):
Looks very good now! Some components could be separated into smaller parts.

packages/webcert/src/feature/certificate/Question/Question.tsx
```diff
@@ -43,6 +46,39 @@ export interface QuestionProps {
...
+`
+
+const CloseLink = styled.h5.attrs({
+  className: 'iu-fs-100 iu-fw-bold iu-lh-body open',
+})`
+  display: none;
+  text-decoration: underline;
+`
+
+const ControlAccordion = styled(Accordion)`
```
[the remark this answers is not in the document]
SopraErik: The accordion for thi input fields nedded some different styling from the accordion fhowing the question description, So just to separate them a bit. Maybe the naming could be better...

PR #583, opened 2022-12-01
INTYGFV-15042: making description field the only mandatry field in mandatory cause of death controls

SopraErik:


SamuelTilly (commented):
Looks good 👍, but I would like some tests added to `validationUtils` that covers this new validation.

packages/common/src/utils/validationUtils.test.ts
```diff
@@ -344,6 +345,59 @@ describe('Validate mandatory rule for uncertain datet', () => {
...
+    },
+    validation: [
+      {
+        type: CertificateDataValidationType.MANDATORY_VALIDATION,
+        questionId: '1',
+        expression: '$dodsorsak',
+      },
+    ],
+    validationErrors: [],
+  }
```
SamuelTilly, line 377: Instead of creating this object, we could use our helper function. For example: `fakeCauseOfDeathElement({ id: '1', value: { text: null } })[1]`

SamuelTilly (approved):

PR #584, opened 2022-12-02
INTYGFV-15051: Refactor expression validation

SamuelTilly:


packages/common/src/utils/validationUtils.test.ts
```diff
@@ -1381,20 +1382,20 @@ describe('autoFillElement', () => {
 describe('Validate expressions with boolean values set to null or undefined should not be showed', () => {
   const element = getBooleanElement()
 
-  it('should return false if selected is undefined for negative expression', () => {
+  it('should return true if selected is undefined for negative expression', () => {
     ;(element.value as ValueBoolean).selected = undefined
     const result = parseExpression('!$dodsdatumSakert', element, CertificateDataValidationType.SHOW_VALIDATION)
-    expect(result).toBe(false)
+    expect(result).toBe(true)
   })
```
SamuelTilly, line 1389: This could potentially introduce a breaking change, but the previous evaluation doesn't make sense to me `!undefined` or `!null` should resolve to `true`

mww-sopra-steria:
Played around in the client locally and all rules tested seems to work fine with one exception. The UncertainDate is displayed initially when creating a draft of either db or doi. It should only be displayed if 'Dödsdatum" has the value 'false', not when it doesn't have a value.

image

SamuelTilly:
> The UncertainDate is displayed initially when creating a draft of either db or doi. It should only be displayed if 'Dödsdatum" has the value 'false', not when it doesn't have a value.

I could change so that it works the same way, or we could update the backend to instead look if value is missing with the expression `exists($dodsdatumSakert) && $dodsdatumSakert` instead of `!$dodsdatumSakert`, this should only affect dodsbevis since it was added here fairly recently here https://github.com/sklintyg/frontend/commit/689efc61b6427178c8c7985a1e3a138ebe3ec629

SamuelTilly:
@mww-sopra-steria I updated the storybook (https://github.com/sklintyg/frontend/pull/584/commits/662e692175113b96890625476219c45ffcd02497) with an example on how backend could be updated for "dodsdatumSakert"

mww-sopra-steria (approved):
Great work!

PR #585, opened 2022-12-02
INTYGFV-15041: This addresses bugs reported in several Jiras

SopraErik:
in addition to this, also the bugs reported in INTYGFV-15045, INTYGFV-15046 and INTYGFV-15047
It also includes the fix from INTYGFV-15042, which has its own PR


packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeathControl.tsx
```diff
@@ -77,6 +81,9 @@ const UeCauseOfDeathControl: React.FC<Props> = ({
     ? (validation.find((v) => v.type === CertificateDataValidationType.TEXT_VALIDATION) as TextValidation)
     : undefined
 
+  const emptyValidationError = validationErrors ? (validationErrors.find((e) => e.type === 'EMPTY') as ValidationError) : undefined
+  const nonEmptyValidationErrors = validationErrors ? validationErrors.filter((e) => e.type !== 'EMPTY') : undefined
+
   const specifications: ConfigUeDropdownItem[] = [{ id: '', label: 'Välj...' }, ...config.specifications]
```
SamuelTilly, line 87: This needs to be changed to `specifications: ConfigUeCodeItem[]` with a code value of `''` for the first option.

packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeathControl.tsx
```diff
@@ -88,7 +95,8 @@ const UeCauseOfDeathControl: React.FC<Props> = ({
   }
 
   const handleSpecificationChange = (code: string) => {
-    onChange({ ...value, specification: { ...value.specification, code } })
+    const specificationId = config.specifications.find((s) => s.code === code)?.id ?? ''
+    onChange({ ...value, specification: { ...value.specification, id: specificationId, code: code } })
```
SamuelTilly, line 99: I just noticed that we are setting `item.id` as value in each option element, so what we are calling `code` here is really "id" right now. Perhaps we can just set an index to the options value and select specification from the config.

```tsx
const handleSpecificationChange = ({ id, code }: ConfigUeCodeItem) => {
  onChange({ ...value, specification: { ...value.specification, id, code } })
}

<Dropdown 
onChange={(event) => {
  handleSpecificationChange(specifications[parseInt(event.currentTarget.value)])
}}
options={specifications.map((item, index) => (
  <option value={index} key={item.id}>
    {item.label}
  </option>
))}
/>
```

SamuelTilly (changes requested):

packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeathControl.tsx
SopraErik: This won't work, since we set the value of the dropdown to specification.code, so mapping the option value to index won't match
image

SamuelTilly: Good point! :)

packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeath.tsx
```diff
@@ -11,18 +19,22 @@ interface Props {
...
   const validationErrors = useSelector(getVisibleValidationErrors(question.id))
   const isShowValidationError = useSelector(getShowValidationErrors)
   const [currentValue, setCurrentValue] = useState<ValueCauseOfDeath>(question.value as ValueCauseOfDeath)
 
   const handleChange = (value: ValueCauseOfDeath) => {
     setCurrentValue(value)
+    let debut = value.debut
+    if (value && !isValid(getValidDate(value.debut.date))) {
+      debut = { id: value.id, type: CertificateDataValueType.DATE, date: '' }
+    }
```
SamuelTilly, line 33: `id` should be `value.debute.id`, and `getValidDate` manipulates the date string before validating so it needs to be sent as the result.

```typescript
const date = getValidDate(value.debut.date)
const debut = { ...value.debut,  date: isValid(date) ? date : '' }
```

packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeathList.tsx
```diff
@@ -90,17 +91,24 @@ const UeCauseOfDeathList: React.FC<Props> = ({ question, disabled }) => {
...
   const handleChange = (value: ValueCauseOfDeath) => {
-    updateList(questionValueList.map((item) => (item.id === value.id ? value : item)))
+    updateList(
+      questionValueList.map((item) => (item.id === value.id ? value : item)),
+      value
+    )
   }
 
-  const updateList = (list: ValueCauseOfDeath[]) => {
+  const updateList = (list: ValueCauseOfDeath[], value?: ValueCauseOfDeath) => {
```
SamuelTilly, line 100: I would prefer that date is corrected for every debut in the list on update, no need for extra value parameter for this function.

```typescript
list.map(({...val, debut}) => {
  const date = getValidDate(debut.date)
  return {...val, debut: { ...debut, date: isValid(date) ? date : '' } }
})
```

packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeath.tsx
SopraErik: Fixed this, but since getValidDate returns Date, it needs to be converted back to string, or we'll get an error

packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeathList.tsx
SopraErik: fixed in similar way as in UeCauseOfDeath.tsx, converting Date back to string.

SamuelTilly (approved):

PR #586, opened 2022-12-05
INTYGFV-15006: Component for medical investigations

UlrikaSopra:


SamuelTilly (changes requested):
Looks good! 👍, only missing some tests for `UeMedicalInvestigationList`

HelenaEdblad:
Fixed the category header when it is an accordion so it looks like the other category headers. Changed from a h5 to h3.

SamuelTilly (approved):

PR #588, opened 2022-12-06
INTYGFV-14970: Adjust heights on input fields

HelenaEdblad:
We decided that the input fields TextInput, Dropdown and DatePicker should be the same height as the TextInput, which is 49.29px (which is 2.956rem).


HelenaEdblad:
We decided that the input fields TextInput, Dropdown and DatePicker should be the same height as the TextInput, which is 49.29px (which is 2.956rem).

packages/webcert/src/feature/certificate/Inputs/UeCauseOfDeath/UeCauseOfDeathList.tsx
```diff
@@ -133,7 +133,11 @@ const UeCauseOfDeathList: React.FC<Props> = ({ question, disabled }) => {
                   validationErrors={validationErrors.filter((v) => v.field.includes(`[${index}]`))}>
                   <DeleteButtonWrapper className="iu-ml-500">
                     {index > 0 && (
-                      <CustomButton disabled={disabled} buttonStyle="secondary" onClick={() => handleDeleteRow(config.id)} height="47px">
+                      <CustomButton
+                        disabled={disabled}
+                        buttonStyle="secondary"
+                        onClick={() => handleDeleteRow(config.id)}
+                        height="2.956rem">
```
SopraErik, line 140: I think this wil probably not be the only case when we will need an in-line button.
I would prefer not to set the height here, but rather maybe a separate control for inline button.

SopraErik (changes requested):

HelenaEdblad:
Changed the height to 48px (3rem) because that makes more sence..

SopraErik (approved):
Great, looking good!

PR #590, opened 2022-12-12
Feature/INTYGFV-15055 and 15056:  modals on DOI when DB exists on other unit 

SopraErik:


HelenaEdblad:
Nice! The only thing I think about is that you could remove feature/ in the title of the PR :)

PR #600, opened 2022-12-15
INTYGFV-15082 & INTYGFV-15084 New layout alternatives for UeCheckboxGroup and UeRadioButtonGroup

SopraErik:


SamuelTilly (changes requested):
image

image

Checkboxes gets a padding top for columns and inline layout, and when inline there is not enough space between checkboxes.

It's the same issue for radio buttons, but the extra spacing is added to the bottom instead of top for inline and columns.

Could we create a shared wrapper component that handles layout?

packages/common/src/types/certificate.ts
```diff
@@ -158,8 +158,16 @@ export interface CheckboxCode {
   disabled?: boolean
 }
 
+export enum ConfigLayout {
+  ROWS = 'ROWS',
+  INLINE = 'INLINE',
+  COLUMN = 'COLUMN',
```
SamuelTilly, line 164: Doesn't look like this is used, perhaps it can be removed?

SopraErik: It is used in backend, so we might get error if not implemented 
I'm changing line 62 of radioGroup, which is what it's actually used for, so it becomes clearer (should be horizontal only has effect on ROWS-layout)
const shouldBeHorizontal = config.layout !== ConfigLayout.COLUMN && radiobuttons.length <= 2

SopraErik: INLINE-layout is meant for very shoer labels, as in drivers licence categories: C, CE, D, DE etc. and then the spacing looks good. Maybe change the storybook to reflect this

ChrilleSigerhed:
After reviewing the implementation of Layout.COLUMN we realised it may be redundant. If we want to display an component in one column we should use Layout.ROWS. This has resulted in that COLUMN has been removed from backend. Illustration of layouts:

ROWS:
-
-
-

COLUMNS:
-  |  -
-  |  -
-  |  -

INLINE:
- - - - - - - - - -

packages/webcert/src/feature/certificate/Inputs/GroupWrappers.tsx
```diff
@@ -0,0 +1,44 @@
+import { ConfigLayout } from '@frontend/common/'
+import styled, { css } from 'styled-components'
+
+interface checkboxesProps {
```
SamuelTilly, line 4: Minor thing, but this should probably be named `GroupWrapperProps` and `checkboxProps` should be  `ItemWrapperProps`

SamuelTilly (approved):

PR #604, opened 2022-12-19
INTYFV-15097: Add recipient to revoke cert modal

HelenaEdblad:


SopraErik (commented):
Should this implement the ner senTo metadata instead?

SopraErik (approved):

PR #618, opened 2022-12-20
INTYGFV-15138: Use new property sentTo in CertificateMetaData to display the recipient for SentStatus and ShowHistory.

mww-sopra-steria:


SopraErik (approved):
Nice work!

PR #623, opened 2022-12-27
INTYGFV-15127, 15131 and 15133: added view-components,  tidied up the inputs folder and refactored UvText

SopraErik:


packages/common/src/types/certificate.ts
```diff
@@ -298,6 +301,23 @@ export interface ConfigUeCauseOfDeathList extends CertificateDataConfig {
   list: ConfigUeCauseOfDeathControl[]
 }
 
+export interface ConfigUeViewText extends CertificateDataConfig {
+  label?: string
```
ChrilleSigerhed, line 305: Should we call this text instead of label? Its what we normaly call this types of "headers" for questions.

packages/common/src/types/certificate.ts
```diff
@@ -298,6 +301,23 @@ export interface ConfigUeCauseOfDeathList extends CertificateDataConfig {
...
+}
+
+export interface ConfigUeViewList extends CertificateDataConfig {
+  label?: string
```
ChrilleSigerhed, line 309: Should we call this text instead of label? Its what we normaly call this types of "headers" for questions.

packages/common/src/types/certificate.ts
```diff
@@ -318,6 +338,10 @@ export enum CertificateDataValueType {
   MEDICAL_INVESTIGATION_LIST = 'MEDICAL_INVESTIGATION_LIST',
   CAUSE_OF_DEATH = 'CAUSE_OF_DEATH',
   CAUSE_OF_DEATH_LIST = 'CAUSE_OF_DEATH_LIST',
+  VIEW_TEXT = 'VIIEW_TEXT',
```
ChrilleSigerhed, line 341: typo, change to VIEW_TEXT.

packages/common/src/types/certificate.ts
```diff
@@ -445,6 +471,26 @@ export interface ValueMedicalInvestigationList extends Value {
   list: ValueMedicalInvestigation[]
 }
 
+export interface ValueViewText extends Value {
+  type: CertificateDataValueType.VIEW_TEXT
+  text: string
+}
+
+export interface ValueViewList extends Value {
+  type: CertificateDataValueType.LIST
```
ChrilleSigerhed, line 480: Should we add the VIEW prefix to list, similar to VIEW_TEXT?

packages/common/src/types/certificate.ts
```diff
@@ -445,6 +471,26 @@ export interface ValueMedicalInvestigationList extends Value {
...
+  list: ValueViewText[]
+}
+
+export interface ValueTextRow extends Value {
+  type: CertificateDataValueType.ROW
```
ChrilleSigerhed, line 485: Should we add the VIEW prefix to row, similar to VIEW_TEXT?

packages/common/src/types/certificate.ts
```diff
@@ -445,6 +471,26 @@ export interface ValueMedicalInvestigationList extends Value {
...
+  columns: ValueText[]
+}
+
+export interface ValueViewTable extends Value {
+  type: CertificateDataValueType.TABLE
```
ChrilleSigerhed, line 490: Should we add the VIEW prefix to table, similar to VIEW_TEXT?

packages/common/src/types/certificate.ts
```diff
@@ -298,6 +301,23 @@ export interface ConfigUeCauseOfDeathList extends CertificateDataConfig {
...
+  label?: string
+}
+
+export interface ConfigUeViewList extends CertificateDataConfig {
+  label?: string
+}
+
+export interface ConfigViewColumn {
+  id: string
+  title: string
```
ChrilleSigerhed, line 314: I suggest we try to follow the same structure that we have in our miro board:
`{`
`text: Sting`
`columns: [
            id: String
            title:  String
        ]`
`}`
This will allow us to create ViewTables with more than one column.

ChrilleSigerhed (changes requested):
Looks great, well done!

I left some suggestions that I think might benefit us. Please take a moment and review the requested changes.

packages/common/src/types/certificate.ts
SopraErik: Text is already there in the interface it extends, this is on top of that and is used in some cases, but definitely not always. So use text

packages/common/src/types/certificate.ts
SopraErik: Text is already there in the interface it extends, this is on top of that and is used in some cases, but definitely not always. So use text

packages/common/src/types/certificate.ts
SopraErik: Will fix

packages/common/src/types/certificate.ts
SopraErik: Definitely

packages/common/src/types/certificate.ts
SopraErik: Definitely

packages/common/src/types/certificate.ts
SopraErik: Definitely

packages/common/src/types/certificate.ts
SopraErik: If you look at ConfigUeViewTable you see that's exactly what we do! (but title as in Miro, instead of text, but maybe text is better

image

ChrilleSigerhed: Awwww... :-) Nice!! 

I think it would make sense to use text instead of title, since thats what we usually call our "headers" 👍

packages/common/src/types/certificate.ts
ChrilleSigerhed: Awsome!

SamuelTilly (approved):

ChrilleSigerhed (approved):

PR #628, opened 2023-01-12
INTYGFV-15075: Issues with validation and dispatch for VisualAcuity

SamuelTilly:


ChrilleSigerhed (changes requested):
Great work, now the values are saved correctly!

We still have some scenarios that we need to fix tho, regarding precision of the float value.

image

1. If no decimal is provided by the user, we should add it after the user leaves the "frame".
2. We dont allow two commas in a row, the correct format is always number, comma, number.

SamuelTilly:
> We still have some scenarios that we need to fix tho, regarding precision of the float value.

Nice catch, should be fixed now

ChrilleSigerhed (approved):
Nice! Approved.

PR #629, opened 2023-01-13
Refactor flaws in typeahead component

SamuelTilly:
Typeahead has a lot of design flaws that makes it difficult to update and reason about. This aims to fix that and simplify TextInput and TextArea component.


SopraErik (changes requested):
Really nice work, 
It all works very well when trying it in the application, on certificates.

but storybook for Diagnoses is not working properly
Focusing on the code column expands all suggestions, but typing doesnt render any results

image

SamuelTilly:
> but storybook for Diagnoses is not working properly Focusing on the code column expands all suggestions, but typing doesnt render any results

That is sadly a limitation of the `UeDiagnosis` component, it calls `resetDiagnosisTypeahead` that clears all diagnosis in the store and the fake store that the storiebook uses has currently no way of answering to `getDiagnosisTypeahead` dispatch. Perhaps we can come up with a better solution in a future improvement, but for now I'll just limit this to typeahead.

SopraErik (approved):

PR #630, opened 2023-01-13
INTYGFV-15124: Fix unable to view signed certificate

SamuelTilly:
Fixes issue with not being able to open FK7804 and AG7804, issue was in `UvDateRange` where `.reverse()` tries to modify immutable `config.list`.

Extra in this PR:
* UvText should print out "Ej angivet" for missing text value
* Cleanup


ChrilleSigerhed (approved):
Approved!

PR #637, opened 2023-01-17
INTYGFV-15178: added dynamic link suport for about certificate

SopraErik:


ChrilleSigerhed (changes requested):
Manually tested and looks great, now we are able to display links in our AboutCertificatePanel!

Would be nice with some unit tests to verify this new behaviour!

packages/webcert/src/feature/certificate/CertificateSidePanel/AboutCertificatePanel.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  it('renders with description', () => {
+    renderComponent()
+    expect(screen.getByText(text1)).toBeInTheDocument()
+    expect(screen.getByText(text2)).toBeInTheDocument()
+    expect(screen.getByText(text3)).toBeInTheDocument()
+  })
+
+  it('renders with description with links', () => {
+    renderComponent()
+    // how can I check that WCDynamic link is called? With correct parameters?
```
UlrikaSopra, line 66: Maybe something like this? 
expect(screen.getByText(text1)).toBeInTheDocument()
expect(screen.getByRole("link", { name: link1 })).toBeInTheDocument()

SopraErik: No, because the link is fetched from backend via the Utils store, so a little bit more complicated than that unfortynately

packages/webcert/src/feature/certificate/CertificateSidePanel/AboutCertificatePanel.tsx
```diff
@@ -55,6 +56,41 @@ const AboutCertificatePanel: React.FC<Props> = ({ headerHeight }) => {
...
+            <WCDynamicLink linkKey={dynamicLinkKey} />
+            <span dangerouslySetInnerHTML={sanitizeText(textAfterLink)}></span>
+          </>
+        )
+      }
+      return <>{returnValue}</>
+    }
+    return text
+  }
+
```
SamuelTilly, line 93: This is also used in CertificateListRow, so we should probably move this out to a utility function. That would also make writing tests for this function easier.

SamuelTilly (changes requested):

packages/webcert/src/feature/certificate/CertificateSidePanel/AboutCertificatePanel.tsx
SopraErik: I've moved it out now, and also updated CertificateListRow accordingly

ChrilleSigerhed (approved):
Nice, well done!

packages/webcert/src/feature/certificate/CertificateSidePanel/AboutCertificatePanel.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+const history = createMemoryHistory()
+
+const text1 = faker.lorem.sentence(10)
+const text2 = faker.lorem.sentence(10)
+const text3 = faker.lorem.sentence(10)
+const link1 = faker.lorem.word()
+const link2 = faker.lorem.word()
+
+const descriptionWithLinks = `${text1} <LINK:${link1}> ${text2} <LINK:${link2}> ${text3}`
+console.log(descriptionWithLinks)
```
SamuelTilly, line 27: console.log should be removed

packages/webcert/src/utils/TextWithDynamicLinks.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+const text2 = faker.lorem.sentence(10)
+const text3 = faker.lorem.sentence(10)
+const link1 = faker.lorem.word()
+const link2 = faker.lorem.word()
+
+const textWithLinks = `${text1} <LINK:${link1}> ${text2} <LINK:${link2}> ${text3}`
+
+const renderComponent = () => {
+  render(
+    <Provider store={testStore}>
```
SamuelTilly, line 26: testStore, Provider etc. is not really necessary for this component since it is not relying on redux functionality.

packages/webcert/src/utils/TextWithDynamicLinks.tsx
```diff
@@ -0,0 +1,48 @@
...
+  text: string
+}
+
+const TextWithDynamicLinks: React.FC<Props> = ({ text }) => {
+  const hasDynamicLink = (text?: string): boolean => {
+    if (!text) {
+      return false
+    }
+    return text.split('<LINK:').length > 1
+  }
```
SamuelTilly, line 15: `const hasDynamicLink = text.includes('<LINK:')`

packages/webcert/src/utils/TextWithDynamicLinks.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const formatText = (text?: string) => {
```
SamuelTilly, line 17: `text` is not optional going in to this component, so no need to have it optional here.

packages/webcert/src/utils/TextWithDynamicLinks.tsx
```diff
@@ -0,0 +1,48 @@
...
+    if (!text) {
+      return ''
+    }
+    const splitText = text.split('<LINK:')
+    if (splitText.length > 1) {
+      let returnValue = (
+        <>
```
SamuelTilly, line 24: fragments are not needed here, it's holding a single <span element :)

```typescript
let returnValue = <span dangerouslySetInnerHTML={sanitizeText(splitText[0])}></span>
```

SamuelTilly (changes requested):
Some minor changes, but looking good.

packages/webcert/src/utils/TextWithDynamicLinks.test.tsx
SamuelTilly: This is not resolved. Provider, testStore and middleware etc. can all be removed from tests for this component :)

packages/webcert/src/utils/TextWithDynamicLinks.tsx
SamuelTilly: Partially resolved, but there is no need for having this as a function. text is never undefined so it could simply be changed to `const hasDynamicLink = text.includes('<LINK:')`

packages/webcert/src/utils/TextWithDynamicLinks.tsx
```diff
@@ -0,0 +1,44 @@
...
+    if (!text) {
+      return false
+    }
+    return text.includes('<LINK:')
+  }
+
+  const formatText = (text: string) => {
+    if (!text) {
+      return ''
+    }
```
SamuelTilly, line 20: `text` is never undefined, so this could be removed as well :)

packages/webcert/src/utils/TextWithDynamicLinks.test.tsx
SopraErik: These are needed, since WCDynamicLink  relies on redux. Tests fails without them

SamuelTilly: Oh, you are correct. Perhaps add a test that populates for example link1 with a dynamic key using `updateDynamicLinks` and test that it shows up.

SopraErik: Added test than link texts are rendered

SamuelTilly (approved):
Very nice! 🥳

PR #643, opened 2023-01-19
INTYGFV-15165: UeYear component

SopraErik:


packages/webcert/src/feature/certificate/UnifiedEdit/UeYear/UeYear.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <DatePickerCustom
+        disabled={disabled}
+        textInputOnChange={handleChange}
+        setDate={handleChange}
+        inputString={yearString}
+        questionId={question.id}
+        max={questionConfig.maxYear?.toString()}
+        min={questionConfig.minYear ? (questionConfig.minYear - 1).toString() : ''}
+        displayValidationErrorOutline={validationErrors.length > 0}
+        onDispatchValidationError={dispatchValidationError}
```
UlrikaSopra, line 61: To forbid future dates use this:  forbidFutureDates={true}

SopraErik: Yes, you can do that, but this is a more dynamic way so that you, in the question config can set a different max year from the current year. So no need to use the forbidFutureDates flag

UlrikaSopra (approved):
I think it looks good, maybe add some styling to the component otherwise super :)
image

PR #645, opened 2023-01-23
INTYGFV-15189: Added control if medical investigation has a value in signed certificate mode

UlrikaSopra:
image


packages/common/src/components/UnifiedView/UvMedicalInvestigationList/UvMedicalInvestigationList.tsx
```diff
@@ -5,33 +5,47 @@ import { Table, TableHeader, TableRow, TableCell } from '../../Table'
...
-        })}
-      </tbody>
-    </Table>
-  </div>
-)
+}> = ({ value, config }) => {
+  let hasCells = false
+  return (
+    <div className={'iu-p-none'}>
+      {config.list.map((medicalConfig) => {
```
SopraErik, line 12: isn't `config.list.forEach((medicalConfig) =>` .... better here, since you're not returning an array?

SopraErik (changes requested):

SopraErik (approved):

PR #646, opened 2023-01-23
INTYGFV-15190: Categories with show/hide validation are not hidden correctly

UlrikaSopra:
Fixed bug Categories with show/hide validation are not hidden correctly.


packages/webcert/src/feature/certificate/Certificate.tsx
```diff
@@ -36,12 +36,11 @@ const Wrapper = styled.div`
   .contentPaperWrapper {
     padding-left: 32px;
     padding-right: 32px;
+    padding-bottom: 20px;
   }
 `
 
 const CategoryWrapper = styled.div`
-  background: #ffffff;
```
ChrilleSigerhed, line 43: We can't remove the background from the CategoryWrapper without affecting the displayed validationerrors in category  funktionsnedsattningar:

image

ChrilleSigerhed (changes requested):

ChrilleSigerhed (changes requested):
Looks great! Now the categories are hidden correctly without impacting the category validation!

Would be nice if we could add some tests that verify this new behaviour! It will make it easier to make changes in the future without causing regression! :-)

ChrilleSigerhed (approved):
Nice!

PR #648, opened 2023-01-24
INTYGFV-15209: date range component

SopraErik:
This also handles [INTYGFV-15208](https://inera.atlassian.net/browse/INTYGFV-15208) - read only view


packages/common/src/components/UnifiedView/UvDateRangeList/UvDateRangeList.tsx
```diff
@@ -0,0 +1,37 @@
...
+  value: ValueDateRangeList
+  config: ConfigUeSickLeavePeriod
+}> = ({ value, config }) => {
+  if (value.list.length > 0 && value.list.some((val) => val.from && val.to)) {
+    return (
+      <Table>
+        <TableHeader>
+          <TableRow>
+            <TableCell>Nedsättningsgrad</TableCell>
+            <TableCell>Från och med</TableCell>
```
UlrikaSopra, line 15: Should we use label from BE? 
image

SopraErik: DateRangeList is an old valuetype used for the sickLeave component, and in this case we don't get any label from config from backend

packages/common/src/components/UnifiedView/UvDateRange/UvDateRange.tsx
```diff
@@ -1,37 +1,25 @@
...
-    )
-  }
-  return null
+const DateRangeWrapper = styled.div`
+  display: inline-block;
+`
+
+export const UvDateRange: React.FC<{ value: ValueDateRange; config: ConfigUeDateRange }> = ({ value, config }) => {
+  return (
+    <>
```
SamuelTilly, line 12: Fragment can be removed.

packages/common/src/utils/faker/fakeCertificateData.ts
```diff
@@ -468,6 +470,33 @@ export const fakeDateElement = (
...
+  fakeDataElement(
+    {
+      ...data,
+      config: {
+        type: ConfigTypes.UE_DATE_RANGE,
+        fromLabel: 'Fr.o.m',
+        toLabel: 't.o.m',
+        ...data?.config,
+      },
+      // value: { type: CertificateDataValueType.DATE, date: '2022-09-29', ...data?.value },
```
SamuelTilly, line 486: comment could be removed.

packages/webcert/src/feature/certificate/Question/QuestionUvResolve.test.tsx
```diff
@@ -412,6 +421,25 @@ export function createQuestionWithMultipleDates(): CertificateDataElement {
...
+    id: '',
+    text: '',
+    type: ConfigTypes.UE_DATE_RANGE,
+    fromLabel: 'Fr.o.m',
+    toLabel: 't.o.m',
+  }
+
+  return createQuestion(value, config)
+}
+
```
SamuelTilly, line 442: This should use the new `fakeDateRangeElement` function instead

packages/webcert/src/feature/certificate/UnifiedEdit/UeDateRange/Styles.tsx
```diff
@@ -0,0 +1,27 @@
...
+`
+export const DateGrid = styled.div`
+  display: grid;
+  align-items: baseline;
+  grid-template-columns: 1fr 1fr;
+  @media (max-width: 1200px) {
+    grid-template-columns: 1fr;
+    grid-gap: 8px;
+  }
+`
```
SamuelTilly, line 27: Move these components into `UeDateRange`  where they are used and remove `Styles.tsx`

SamuelTilly (changes requested):
There are some minor changes, but overall it looks good. However, I wonder if we could queue this PR after the changes to clientValidation since that would significantly simplify this component, and we could avoid having too many merge conflicts. https://github.com/sklintyg/frontend/pull/644

ChrilleSigerhed (changes requested):
Is it possible to only highlight the date that is missing an value?

image

SopraErik:
> Is it possible to only highlight the date that is missing an value when validating?
> 
> image

I will take a renewed look at this once 15177 is merged, since that includes som major updates to how det datepicker works

SopraErik:
> > Is it possible to only highlight the date that is missing an value when validating?
> > image
> 
> I will take a renewed look at this once 15177 is merged, since that includes som major updates to how det datepicker works

This is fixed now

SamuelTilly (approved):
Very nice!

ChrilleSigerhed (approved):
Well done!

PR #653, opened 2023-01-26
INTYGFV-15211: Added tests for date range expression used in new component ue_date_range.

ChrilleSigerhed:


packages/common/src/utils/test/certificateTestUtil.ts
```diff
@@ -204,7 +204,28 @@ export const getSickLeavePeriodElement = (): CertificateDataElement => {
...
+export const getDateRangeElement = (): CertificateDataElement => {
+  return {
+    id: '1.4',
+    parent: '1.1',
+    index: 1,
+    visible: true,
+    mandatory: false,
+    readOnly: false,
+    config: {
+      type: ConfigTypes.UE_DATE,
```
SopraErik, line 216: Shouldn't ConfigType be UE_DATE_RANGE?

SopraErik (changes requested):

packages/common/src/utils/test/certificateTestUtil.ts
ChrilleSigerhed: Yes, I would prefer that aswell. I'll go ahead and add it!

SopraErik (approved):

PR #658, opened 2023-01-30
INTYGFV-15180: After test, added same functionality to show history 

SopraErik:
i.e. remove available to patioent for death and cause of death certificates


packages/webcert/src/feature/certificate/CertificateHeader/ShowHistory.tsx
```diff
@@ -46,7 +46,8 @@ const ShowHistory: React.FC<Props> = ({ historyEntries, certificateMetadata }) =
       case CertificateEventType.SIGNED:
         return 'Intyget är signerat'
       case CertificateEventType.AVAILABLE_FOR_PATIENT:
-        return 'Intyget är tillgängligt för patienten'
+        if (certificateMetadata.type === 'db' || certificateMetadata.type === 'doi') return
+        else return 'Intyget är tillgängligt för patienten'
```
SamuelTilly, line 50: Should this perhaps be handled by backend instead, so that "AVAILABLE_FOR_PATIENT" never gets received for these types of certificates?

SopraErik:
This will be handled by backend instead. Closing PR

PR #663, opened 2023-01-31
INTYGFV-15204: Integer component

UlrikaSopra:
Integer component and fixing merge problem


packages/common/src/types/certificate.ts
```diff
@@ -98,6 +98,7 @@ export enum ConfigTypes {
   UE_VIEW_LIST = 'UE_VIEW_LIST',
   UE_VIEW_TABLE = 'UE_VIEW_TABLE',
   UE_YEAR = 'UE_YEAR',
+  UE_INTEGER = 'UE_INTEGER',
 }
 
 export type CertificateDataConfigType =
```
SopraErik, line 104: You should probly add ConfigUeInteger here

packages/common/src/utils/faker/fakeCertificateData.ts
```diff
@@ -471,6 +473,22 @@ export const fakeYearElement = (
...
   )
 
+export const fakeIntegerElement = (
+  data?: PartialCertificateDataElement<ConfigUeInteger, ValueInteger>,
+  children?: CertificateData[]
+): CertificateData =>
+  fakeDataElement(
+    {
+      ...data,
+      config: {
```
SopraErik, line 483: You should use fakeCertificateConfig here

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
```diff
@@ -0,0 +1,107 @@
...
+    isNumberValid(inputValue)
+    dispatch(
+      updateCertificateDataElement({
+        ...question,
+        value: { ...questionValue, value: inputValue },
+      })
+    )
+  }
+
+  const isNumberValid = (number: string): ValidationError | undefined => {
```
SopraErik, line 71: What does this do? Are you really using the returned object? 
0 and 100 should not be hard coded, you should use min and max values from config

SopraErik (changes requested):

packages/common/src/types/certificate.ts
```diff
@@ -412,6 +413,13 @@ export interface ConfigUeYear extends CertificateDataConfig {
   maxYear?: number
 }
 
+export interface ConfigUeInteger extends CertificateDataConfig {
+  id: string
```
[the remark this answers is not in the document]
UlrikaSopra: Not according to our data model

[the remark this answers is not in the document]
SopraErik: Some changes have been made here, look for example how the othe configs are specified now
image

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
UlrikaSopra: Maybe no longer needed? I will remove it :)

SopraErik (changes requested):
Entering a '-' in another position than the 1st causes an error
Ideally a '-' in another position than the 1st should be handled like entering a non numeric character.

image

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  const handleNumberOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const inputValue = event.target.value
+
+    if (/^[a-zA-Z]+$/.test(inputValue)) {
+      return null
+    }
+    setNumber(event.target.value.replace(/[^0-9-]/g, ''))
```
SopraErik, line 50: event.target.value could be replaced with inputValue

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
```diff
@@ -42,12 +42,14 @@ const UeInteger: React.FC<Props> = ({ question, disabled }) => {
...
   const handleNumberOnChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const inputValue = event.target.value
-
+    let inputValue = event.target.value
     if (/^[a-zA-Z]+$/.test(inputValue)) {
-      return null
+      return
+    }
+    if (inputValue.indexOf('-') > 0) {
+      inputValue = inputValue.replace('-', '')
```
SopraErik, line 50: This will only replace the first occurance of '-', if someone pastes in a number with more than onr '-' you'll get the same error.  you'll probably need something like replace(/-/g,'') instead.

SopraErik (changes requested):

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
SamuelTilly: The above regex `/^[a-zA-Z]+$/.test(inputValue)` could be changed to `/[^\d]/.test(inputValue)` to only allow numerical.

UlrikaSopra: hmm but we also want to handle negative numbers

SamuelTilly: Alright, we can include that too, but it becomes a bit more readable if we flip the expression `!/^-?[\d]+$/.test(inputValue)`

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <>
+      <Wrapper>
+        <StyledTextInput
+          disabled={disabled}
+          value={number}
+          id={questionConfig.id}
+          onChange={handleNumberOnChange}
+          hasValidationError={validationErrors.length > 0}
+          limit={3}
```
SopraErik, line 62: limit should be set from allowed min och max values from config. 
something like config.max && config.min ? max(config.min.toString().length, config.max.toString().length) : 16,

As it is now if max is 1000 or min is -100 you can't enter these

SopraErik (changes requested):

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
SopraErik: Alternatively don't set at all, since the validation is made server side

SopraErik (approved):

PR #665, opened 2023-02-01
adjusted styling for accordion header and mandatory icon when headeris h5

SopraErik:


SamuelTilly (approved):
Nicely done! looks much better now 👍

PR #666, opened 2023-02-01
INTYGFV-15240: Added shadow when focusing on input with error

mhornfeldt:


SamuelTilly (changes requested):
Looks good for TextInput, but we should probably fix TextArea as well.

image

Missing shadow and background changes when gaining focus.

SamuelTilly:
I added a bunch of storiebooks for the shared components to make this a bit easier.

SopraErik (changes requested):
Text area gets pink background on focus even if there are no validation errors
image

SamuelTilly (approved):

SamuelTilly (approved):

SopraErik (approved):

PR #668, opened 2023-02-02
Prevent history from reaching reducers

SamuelTilly:
We have an issue with `history` object reaching reducers that creates a lot of noise when running tests. This PR introduces https://github.com/supasate/connected-react-router that replaces all instances of history in our middleware.

https://redux.js.org/faq/actions#why-should-type-be-a-string-or-at-least-serializable-why-should-my-action-types-be-constants


mhornfeldt (approved):

SopraErik (approved):
Really nice work!

PR #674, opened 2023-02-06
Bug/INTYGFV-15200: Disallow input of numbers that start with 0 and more characters

UlrikaSopra:


SopraErik (changes requested):
If I start to type in a number starting with 0, no further entry is possible.
Also, I can enter -01, but gets an error

image

image

packages/webcert/src/feature/certificate/UnifiedEdit/UeInteger/UeInteger.tsx
```diff
@@ -26,7 +26,7 @@ const UeInteger: React.FC<Props> = ({ question, disabled }) => {
   const dispatch = useDispatch()
   const questionValue = question.value as ValueInteger
   const questionConfig = question.config as ConfigUeInteger
-  const [number, setNumber] = useState(questionValue.value?.toString() ?? '')
+  const [number, setNumber] = useState(questionValue.value?.toString() ?? undefined)
```
SopraErik, line 29: Maybe a better idea would be to have the local state numeric rather than string?

SopraErik (approved):
Now it seems to work as intended :)

UlrikaSopra:
> Now it seems to work as intended :)

Finally! :D

PR #676, opened 2023-02-06
INTYGFV-15245: Fix issue with update loop for ueIcf and UeDropdown

SamuelTilly:


SopraErik (approved):
Nice work! 
This also seem to have solved some of the reported problems regarding fk7804

PR #677, opened 2023-02-06
INTYGFV-14802: split up SearchAndCreatePage to two pages, and updated url:s etc in menus, buttons, and resourceAccess

SopraErik:


SopraErik:
Need to update some tests etc.

SamuelTilly (commented):
Nice work!, good to have the pages separated. But it seems the error in INTYGFV-14802 is not resolved?

SamuelTilly:
> > Nice work!, good to have the pages separated. But it seems the error described in INTYGFV-14802 is not resolved?
> 
> As far as I cen tell, it now works the same way as in the angular client.
> 
> Select a patient Change patient press back => your get, once again, the search page for patient.
> 
> Or what are you thinking about

Nevermind, I had a local bug/INTYGFV-14802 branch for when we looked at this last year. After updating, it works as expected :D

SamuelTilly (approved):

PR #679, opened 2023-02-07
INTYGFV-15256: Updated read only view for medical investigation list

mhornfeldt:


SopraErik (approved):

packages/common/src/utils/faker/fakeCertificateData.ts
```diff
@@ -255,6 +255,57 @@ export const fakeMedicalInvestigationListElement = fakeDataElement<ConfigUeMedic
   }
 )
 
+export const fakeEmptyMedicalInvestigationListElement = fakeDataElement<ConfigUeMedicalInvestigationList, ValueMedicalInvestigationList>(
```
SamuelTilly, line 258: How does this differ from `fakeMedicalInvestigationListElement` ?

SamuelTilly: looks like it's using adding an empty value list, so you could replace this with `fakeMedicalInvestigationListElement({value: { list: [] }})`

SamuelTilly (changes requested):

packages/common/src/utils/faker/fakeCertificateData.ts
mhornfeldt: Good suggestion! Fixed it now.

SamuelTilly (approved):

PR #684, opened 2023-02-08
INTYGFV-15253: added padding to bottom of last item in question side panel.

SopraErik:
For smaller screens sometimes the bottom of the sidepannel hides behind its footer, making buttons unrechable anf text unreadable.
By adding a padding to the bottom of the last item in the sidepannel above its footer, we elements getting hidden.


UlrikaSopra (approved):
Looks good!

SopraErik:
> I'm having issues duplicating this issue in master, instead I get a bit too much padding :/

It tends to be fror smaller window sizes, with a height around 1000 px. But it van be a bit difficult to dublicate, since it depends on how much info is in the side panel.

PR #686, opened 2023-02-09
INTYGFV-15258: Fixed bug with lines becoming mandatory

UlrikaSopra:


packages/webcert/src/feature/certificate/UnifiedEdit/UeMedicalInvestigation/UeMedicalInvestigation.tsx
```diff
@@ -51,12 +51,16 @@ const UeMedicalInvestigation: React.FC<Props> = ({ disabled, config, value, vali
...
+
+    informationSource.text = text || null
+
     onChange({
       ...value,
       informationSource: {
         ...value.informationSource,
         id: config.informationSourceId,
-        text,
+        text: informationSource.text,
```
mhornfeldt, line 63: Wouldn't it be the same to write text || null in the onChange object instead of assigning it to informationSource.text?

packages/webcert/src/feature/certificate/UnifiedEdit/UeMedicalInvestigation/UeMedicalInvestigation.tsx
```diff
@@ -51,12 +51,16 @@ const UeMedicalInvestigation: React.FC<Props> = ({ disabled, config, value, vali
   }
 
   const handleInformationSourceChange = (text: string) => {
+    const informationSource = { ...value.informationSource }
```
mhornfeldt, line 54: Test should be added to verify that the value is set to null if the text is empty and that the value is not set to null if the text is not empty.

packages/webcert/src/feature/certificate/UnifiedEdit/UeMedicalInvestigation/UeMedicalInvestigationList.tsx
```diff
@@ -83,7 +83,7 @@ const UeMedicalInvestigationList: React.FC<Props> = ({ question, disabled }) =>
               <UeMedicalInvestigation
                 config={config}
                 disabled={disabled}
-                error={index === 0 && validationErrors.length > 0}
+                error={index === 0 && validationErrors.length === 1}
```
mhornfeldt, line 86: Test should be added

mhornfeldt (changes requested):

packages/webcert/src/feature/certificate/UnifiedEdit/UeMedicalInvestigation/UeMedicalInvestigation.tsx
UlrikaSopra: Good idea!

mhornfeldt (approved):

PR #687, opened 2023-02-09
Fix import issues between common and webcert

SamuelTilly:
This aims to fix the following issues:

`common` should not import anything from `@frontend/webcert` this would create recursive dependency
`common` should export images and components that `webcert` depends on.
`webcert` should not import components from `@frontend/common/src/*`
`webcert` should not import images from `@frontend/common/src/*`

These are some issues that prevent us from having correct auto-import, and using go-to-source in vscode.
This will make it easier for `common` to become a proper package for `webcert` and future applications to use.


packages/webcert/src/components/error/modals/ComplementaryCertificateExists.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
```
SopraErik, line 1: Thare have been some changes in these error modals in the master branch (by Michaela) that you need to make sure they are not overwritten by the merge
https://github.com/sklintyg/frontend/pull/685/

SopraErik (changes requested):
Looks really good, but I think there might be some changes regarding error modals merged into master that aren't propperly merged into this branch, so this needs to be addressed, sp changes aren't undone

packages/webcert/src/components/error/modals/ComplementaryCertificateExists.tsx
SamuelTilly: Good catch, I fixed that and manually went through every file to see that I haven't missed anything else. Should be good now :)

SopraErik (approved):

PR #696, opened 2023-02-14
INTYGFV-15264: changed logout method to post

SopraErik:
Inspired by how @mww-sopra-steria  implemented it in Angular
When running locally we'll get an error since CSRF is not implemented locally, as I understand?
But hopefully this should work in our other environments

Might be possible to some pure js-solution, but @mww-sopra-steria  mentioned this should be done as form submit, if I understnd correctly, not sure why though

Maybe the logout methods in redux needs some updating too?


SamuelTilly (approved):
@SopraErik, I think you were on the right track, I made some minor changes to the component to skip the need for a `useEffect` and added `/logout` to the proxy so that it works locally.

After the browser visits /logout we are redirected to /welcome.html (from the old client) so I added that as well in our App.tsx.

PR #700, opened 2023-02-14
INTYGFV-15258: Fixed error displays for UeMedicalInvestigationList. This makes sure that we only display errors if field match. Added handler for when we only have one error.

ChrilleSigerhed:


UlrikaSopra (changes requested):
The field is not highlighted in red, because of error &&
image
Before: 
image

SopraErik (approved):

UlrikaSopra (approved):
Great work!

PR #703, opened 2023-02-16
INTYGFV-14669: positioned page size filter to the right, just above table, next to caption

SopraErik:
…caption


UlrikaSopra (approved):
Nice!

PR #708, opened 2023-02-17
INTYGFV-15287: change color of icon and text for error statuses

SopraErik:


UlrikaSopra (changes requested):
The link has the old color 
image

UlrikaSopra (approved):

PR #711, opened 2023-02-17
INTYGFV-15050: Remove MAX & MIN date validation and replace with config

SamuelTilly:


mhornfeldt (changes requested):
Doesn't max and min need to be set for UeCauseOfDeath, UeMedicalnvestigation, also?

mhornfeldt (approved):

PR #712, opened 2023-02-17
INTYFFV-14825: align question buttons to the right, so forward button is always to the right, even when no other buttons present

SopraErik:


UlrikaSopra (approved):
Nice!

PR #715, opened 2023-02-20
INTYGFV-15269: Upgrade build system, react and dependencies

SamuelTilly:
This aims to replace CRA and jest with vite, vites and fix dependency issues between common and webcert. This introduces faster builds, tests and turbo for managing packages within the repository.

https://vitejs.dev/
https://vitest.dev/
https://turbo.build/

I have removed storybook temporarily to better focus on getting everything else working. Storybook can be re-introduced easily later, but could cause a lot of noise when trying to deprecate CRA.


SamuelTilly:
302 redirect is broken when using the welcome screen for now https://github.com/vitejs/vite/issues/9520
Simply load remove `welcome` from the address works, but could be a bit annoying.

SopraErik (approved):

PR #718, opened 2023-02-20
INTYGFV-15307: First suggestion should be chosen when typeahead opens

mhornfeldt:


SopraErik (approved):

packages/common/src/components/Inputs/Typeahead.tsx
```diff
@@ -77,6 +77,11 @@ const Typeahead = React.forwardRef<HTMLInputElement, Props>(
       onClose && onClose()
     }, [onClose])
 
+    const handleOpen = () => {
+      setOpen(true)
+      setCursor(0)
```
SamuelTilly, line 82: Perhaps it could use the `setCursor` in `handleClose` to reset the initial state. `suggestions.length > 0 ? 0 : -1`

PR #721, opened 2023-02-21
INTYGFV-15288: Updated styling of validation text

mhornfeldt:


mhornfeldt:
image

UlrikaSopra:
Perhaps we also need to add some space in between
image

SopraErik (changes requested):
spacing below text fields needs to be reduced
Inera design:
image

spacing beneath radio- and checkboxes needs to be increased 
Inera design:
image

Since we don't follow the new Inera design yet we will have to implement all of these design changes. Therefore I don't think it's a good idea to have different spacing depending on what type of input element there is because this will require a lot of changes. If we forget to add these changes to some components it might start to look uneven. If we want to increase the spacing between the input element and the validation we should have the same for all components. Maybe you can discuss this at your next UX-meeting and create a new jira when you have decided?

mhornfeldt:
> spacing below text fields needs to be reduced Inera design: image
> 
> spacing beneath radio- and checkboxes needs to be increased Inera design: image

Since we don't follow the new Inera design yet we will have to implement all of these design changes. Therefore I don't think it's a good idea to have different spacing depending on what type of input element there is because this will require a lot of changes. If we forget to add these changes to some components it might start to look uneven. If we want to increase the spacing between the input element and the validation we should have the same for all components. Maybe you can discuss this at your next UX-meeting and create a new jira when you have decided?

SopraErik (approved):
I'll approve this, but have commented in the Jira, so we might get it back as a result of UX-decisions

PR #722, opened 2023-02-21
bug/INTYGFV-15298: Changed creation date to signed certificate date in signed certificate

UlrikaSopra:


packages/webcert/src/feature/certificate/CareUnit/UvCareUnitAddress.tsx
```diff
@@ -36,6 +36,8 @@ const CareUnitAddress = styled.section`
 
 const UvCareUnitAddress: React.FC = () => {
   const metadata = useSelector(getCertificateMetaData, _.isEqual)
+  const signedCertificate = useSelector(getCertificateEvents, _.isEqual)
+  const signedCertificateDate = signedCertificate.reduce((acc, obj) => (obj && obj.timestamp ? obj.timestamp : acc), '')
```
SopraErik, line 41: Don't you need to check that the event type is a signing event? As I read it, this will just return the latest event irrespective of what type of event it is. Could be a sent event, a revoke or replace event, or something else.  Correct me if I'm wrong.

SopraErik (changes requested):

SopraErik (approved):
Nice work!

PR #723, opened 2023-02-21
INTYGFV-15286: Highlight the whole element when blundering

UlrikaSopra:
Same behavior on both of these elements
image


packages/webcert/src/components/header/WebcertHeaderUser.tsx
```diff
@@ -16,6 +16,9 @@ const ExpandableBoxWrapper = styled.div<Props>`
   display: flex;
   align-items: center;
   cursor: ${(props) => (props.changeLinkPointer ? 'pointer' : 'default')};
+  &:focus {
+    outline: 2px solid #a1958a;
+  }
```
SopraErik, line 21: This should only be present when it's actually a clickable element (i.e when it's the private practitioner portal) 
The same fot the click and keydown events, they should fire only if it's the private practitioner portal

SopraErik (changes requested):

SopraErik (approved):

PR #724, opened 2023-02-22
INTYGFV-15311: changed focus of text inputs to be a border rather than shadow

SopraErik:
This included som restructuring of DatePickerCustom to achieve a desired focus behaviour


UlrikaSopra (approved):
Awesome!

PR #725, opened 2023-02-22
INTYGVFV-15310: Align chevron icon in all certificates

UlrikaSopra:


SopraErik (changes requested):
The horizontal alignment looks fine now,

But the chevron in the table header needs to come up abit, so the vertical alignment with the text is similar to how it is in question headings

image

image

image

image

SopraErik:
> The horizontal alignment looks fine now,
> 
> But the chevron in the table header needs to come up abit, so the vertical alignment with the text is similar to how it is in question headings
> 
> image
> 
> image
> 
> image
> 
> image

But wait until https://github.com/sklintyg/frontend/pull/727 since the chevron icons are replaced here, so it might look sligtly different then

SopraErik (changes requested):
Still a difference between table header chevron and normal question chevron

image

UlrikaSopra:
image

SopraErik (approved):

PR #727, opened 2023-02-23
INTYGFV-14625: Replace fontawesome with inera icons

SamuelTilly:


packages/webcert/src/feature/certificate/Buttons/ForwardCertificateButton.tsx
```diff
@@ -60,7 +58,7 @@ const ForwardCertificateButton: React.FC<Props> = ({
         onClick={handleEmailSend}
         tooltip={description}
         data-testid="forward-certificate-button"
-        startIcon={<FontAwesomeIcon size="lg" icon={faShare} />}>
```
SopraErik, line 63: The forward arrow points in the wrong direction
Maybe use transform: scaleX(-1)  to reverse it
Also, the vertical alignment looks a bit off

image

SopraErik (changes requested):

packages/webcert/src/feature/certificate/Buttons/ForwardCertificateButton.tsx
SamuelTilly: Good catch, fixed this. I think the alignment looks a bit better when flipped
image

SopraErik (approved):
Nice job!

SamuelTilly:
Something strange happening with eslint for this branch

PR #731, opened 2023-02-23
INTYGFV-15288: Updated spacing for validation error text

mhornfeldt:


SamuelTilly (changes requested):
I would rather see that `QuestionValidationTexts` always has the same spacing. Perhaps the leading element needs to be updated instead with margin-bottom.

mhornfeldt:
> I would rather see that `QuestionValidationTexts` always has the same spacing. Perhaps the leading element needs to be updated instead with margin-bottom.

What do you mean with the leading element? I don't know if I understand the comment but if you mean why only some components have the spacing prop activated it's because of the requirement. The requirement says to have spacing between the error and the component for example for radio buttons and checkboxes. For textboxes the requirement was to not have any spacing, here I've added a small spacing so the text is not covered by the focus border.

SamuelTilly:
> The requirement says to have spacing between the error and the component for example for radio buttons and checkboxes. For textboxes the requirement was to not have any spacing, here I've added a small spacing so the text is not covered by the focus border.

Alright, in that case the leading elements I'm referring too is radio and checkboxes, whatever comes before `QuestionValidationTexts`. We could wrap those in an element that have something like.

```
:not(:last-child) {
 margin-bottom: 1rem;
}
```
Or adding "iu-mb-300" when `validationErrors.length > 0` etc.

I like to avoid `margin-top` as much as possible in the certificate, since it easily leads to inconsistencies.

mhornfeldt:
> > The requirement says to have spacing between the error and the component for example for radio buttons and checkboxes. For textboxes the requirement was to not have any spacing, here I've added a small spacing so the text is not covered by the focus border.
> 
> Alright, in that case the leading elements I'm referring too is radio and checkboxes, whatever comes before `QuestionValidationTexts`. We could wrap those in an element that have something like.
> 
> ```
> :not(:last-child) {
>  margin-bottom: 1rem;
> }
> ```
> 
> Or adding "iu-mb-300" when `validationErrors.length > 0` etc.
> 
> I like to avoid `margin-top` as much as possible in the certificate, since it easily leads to inconsistencies.

I can change it but I think this will lead to duplicated code and more inconsistencies since it requires us to add the same styling to existing and new components. By doing this we are indirectly creating a px based requirement since we are adding the same spacing to several different places rather than adding it in one place.

SamuelTilly:
> I can change it but I think this will lead to duplicated code and more inconsistencies since it requires us to add the same styling to existing and new components. By doing this we are indirectly creating a px based requirement since we are adding the same spacing to several different places rather than adding it in one place.

We can skip "iu-mt-100", that translates to roughly 1px, so it's not really useful.

This doesn't look right, perhaps we could skip this?
image

This looks right because of the spacing between each radio button
image

Perhaps all we need to is change is `GroupWrapper` to have `margin-bottom: 0.625rem;`

```
export const GroupWrapper = styled.div<Props>`
  :not(:last-child) {
    margin-bottom: 0.625rem;
  }
  ${(props) => {
    if (props.layout === ConfigLayout.COLUMNS) {
      return css`
        display: grid;
        grid-template-columns: 1fr 1fr;
        row-gap: 0.9375rem;
      `
    }
  }}
`
```

`UeRadioGroupOptionalDropdown` could also use GroupWrapper, it's basically the same as UeCheckboxGroup and UeRadioGroup anyway.

UlrikaSopra:
the required sign is still displayed on this radio button, not sure if it has to do with your changes but it works in devtestimage

mhornfeldt:
> the required sign is still displayed on this radio button, not sure if it has to do with your changes but it works in devtestimage

Styling of the error messages doesn't affect the validation rules.

UlrikaSopra (approved):
sry I haven't built common and webcert for a long time, I think it looks good now!

PR #745, opened 2023-03-02
INTYGFV-15402: Lists should show error if user does not have access t…

mhornfeldt:
…o url


SamuelTilly:
Looks good! I have a minor refactoring request for `ResourceAccess`. Instead of having the `resourceAccessMap`, it could require a `ResourceLinkType` property. Each page could supply the correct resource link it needs.

So that for CreatePageWithRedirect it would be `<ResourceAccess linkType={ResourceLinkType.ACCESS_SEARCH_CREATE_PAGE}>`

SamuelTilly (approved):

PR #754, opened 2023-03-06
Re-enable test-coverage report

SamuelTilly:


packages/webcert/src/feature/certificate/UnifiedEdit/UeDateRange/UeDateRange.test.tsx
```diff
@@ -46,7 +46,7 @@ describe('Date range picker', () => {
   })
 
   describe('Validation error', () => {
-    it('shows date range error when end date is before start date', () => {
+    it.skip('shows date range error when end date is before start date', () => {
```
mhornfeldt, line 49: Why do we skip this?

mhornfeldt (approved):

packages/webcert/src/feature/certificate/UnifiedEdit/UeDateRange/UeDateRange.test.tsx
SamuelTilly: Good catch, this test was not working due to the use of `setTimeout`, but I can fix it instead of adding .skip.

SamuelTilly: Nevermind, I removed the test instead. It was trying to test a feature that has been moved to `getClientValidationError` and should no longer be handled by `UeDateRange`

PR #761, opened 2023-03-07
INTYGFV-15354: Removed empty date from getClientValidation.test.ts. I…

ChrilleSigerhed:
…nstead the empty validation gets handled by backend validation.


SamuelTilly:
This will cause some conflict with https://github.com/sklintyg/frontend/pull/749, will theese changes be needed after those are merged?

ChrilleSigerhed:
> 

The changes will still be needed, even tho we might look at another type of solution. The issue with returning an empty_date client validation is that the value of the to/from date is not updated if we keep this logic. This is from handleUpdateCertificateDataElement.

```javascript
const clientValidationErrors = getClientValidationErrors(action.payload)
    dispatch(setValidationErrorsForQuestion({ questionId: action.payload.id, validationErrors: clientValidationErrors }))

    if (clientValidationErrors.length === 0) {
      dispatch(setCertificateDataElement(action.payload))
      dispatch(validateCertificateInFrontEnd(action.payload))
      dispatch(validateCertificate(certificate))
      dispatch(autoSaveCertificate(certificate))
    }
```

ChrilleSigerhed:
I guess this logic was added cause FK7804 & AG7804 contains a list of DateRangeValues.

image

It might be a good idea to look into if we are able to use the field thats provided and map it to the component that has the errors?

image

ChrilleSigerhed:
Updated backend so that we provide a message and added fields for AG7804 & FK7804 so error-management, result:

image

PR #765, opened 2023-03-09
INTYGFV-15410: Change padding on accordian header

UlrikaSopra:


packages/common/src/components/utils/AccordionHeader.tsx
```diff
@@ -2,12 +2,13 @@ import styled from 'styled-components'
...
 const AccordionHeader = styled.summary.attrs({
   tabIndex: 0,
-  className: 'ic-expandable-button  ic-inner iu-focus ic-expandable-button--chevron iu-fs-400 iu-mb-300',
+  className: 'ic-expandable-button iu-focus ic-expandable-button--chevron iu-fs-400 iu-mb-300',
 })`
   justify-content: flex-start;
   align-items: flex-start;
   padding-left: 0 !important;
   margin-bottom: 0.625rem !important;
+  padding-right: 20px !important;
```
SamuelTilly, line 11: image

I think we might need another 10px here

SamuelTilly (changes requested):

packages/common/src/components/utils/AccordionHeader.tsx
UlrikaSopra: Have talked to Rebecka about this and 20px is standard between text and chevron in the expandable component according to her. https://inera.atlassian.net/browse/INTYGFV-15410

SamuelTilly: So this should be changed to 40px then? To account for the width of the icon, itself :)

SamuelTilly (approved):

PR #766, opened 2023-03-09
INTYGFV-15397: Added SRS recommendations to existing panel

mhornfeldt:


packages/common/src/components/utils/ExpandableTextWithTitle.tsx
```diff
@@ -0,0 +1,53 @@
...
+    setExpanded(false)
+  }, [text])
+
+  return (
+    <Root>
+      <TitleBox>
+        <p>{title}</p>
+        <div>
+          <StyledLink href="#" onClick={() => setExpanded(!expanded)}>
+            <p>{expanded ? 'Visa mindre' : 'Visa mer'}</p>
```
SamuelTilly, line 41: Remove paragraph element form link/button

packages/common/src/components/utils/ExpandableTextWithTitle.tsx
```diff
@@ -0,0 +1,53 @@
...
+  useEffect(() => {
+    setExpanded(false)
+  }, [text])
+
+  return (
+    <Root>
+      <TitleBox>
+        <p>{title}</p>
+        <div>
+          <StyledLink href="#" onClick={() => setExpanded(!expanded)}>
```
SamuelTilly, line 40: Should be a button element instead of anchor. It's currently disabled in common for legacy reasons, but we might want to enable this rule: [anchor-is-valid](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/93f78856655696a55309440593e0948c6fb96134/docs/rules/anchor-is-valid.md)

SamuelTilly (changes requested):

SamuelTilly (approved):
Awesome!

PR #772, opened 2023-03-14
Add form components to ids-react-ts

SamuelTilly:


ChrilleSigerhed (approved):
Cool, well done! Im assuming these are reusable components, and that we implement the needed methods in designated class?

SamuelTilly:
> Cool, well done! Im assuming these are reusable components, and that we implement the needed methods in designated class?

Almost, we are simply wrapping already defined Lit components in the `@inera/ids-core` package into React components using `@lit-labs/react` so that we can use them inside our React applications. `@inera/ids-react` does the same thing, but has issues with the exported result at the moment. 

Once `@inera/ids-react` is fixed, we can replace it with this package

PR #774, opened 2023-03-14
INTYGFV-15413: Update value when the component goes from disabled…

UlrikaSopra:
… to enabled


ChrilleSigerhed (commented):
Works great, nice! One question.

In angular the value gets changed back to "Välj tidsperiod" when disabled.

image

Is this something we can achive in react aswell or should we consider this solution "good enought"?

ChrilleSigerhed (approved):
Well done!

PR #788, opened 2023-03-20
Move CertificateDeletedModal to CommonLayout

SamuelTilly:


mhornfeldt:
I can't test this locally but doesn't this fix introduce a bug where the deleted modal is shown if you create a certificate from the create page and delete it?

SamuelTilly:
> I can't test this locally but doesn't this fix introduce a bug where the deleted modal is shown if you create a certificate from the create page and delete it?

From what I can tell, `updateRoutedFromDeletedCertificate` is only set to true in `handleDeleteCertificateSuccess` when the certificate has a parent relation, so I think we are safe but let's test it out before merge

mhornfeldt (approved):

PR #802, opened 2023-03-29
INTYGFV-15340: First implementation of current sickleaves table

mhornfeldt:


mhornfeldt:
Problem som jag tror är kopplat till Inera UX:
- Hjälpikonen hoppar när man checkar in/ut checkbox
- Ikonen i Dölj sökfilter har fel storlek när man först laddar sidan
- Tooltipen i TableHeader följer table-stylingen och visar inte all text

SamuelTilly (changes requested):
Very nice work, there are however quite a few places I would like to make suggestions, which isn't surprising with a larger PR. I noticed the list of changes is getting rather large, so I'll just make a PR instead with the changes to this branch.

SamuelTilly (approved):

PR #808, opened 2023-04-04
INTYGFV-15491: Navigation between sickleaves and patient view

SamuelTilly:


SamuelTilly:
> Nice!! I see you've removed the tab indexes in a few places, aren't they necessary?

They should not be used on non-interactive elements, it becomes difficult to navigate the table with tab when a lot of minor elements gain focus.

PR #809, opened 2023-04-04
INTYGFV-15372: Implemented doctor and sickleave length filter

mhornfeldt:


apps/rehabstod/src/store/api.ts
```diff
@@ -55,20 +55,27 @@ export const api = createApi({
...
-      query: () => ({
+    getSickLeaves: builder.mutation<SickLeaveInfo[], ActiveSickLeavesRequest>({
+      query: (request) => ({
         url: 'sickleaves/active',
         method: 'POST',
+        body: request,
       }),
       transformResponse: (response: { content: SickLeaveInfo[] }) => response.content,
     }),
+    getPopulatedFilters: builder.query<PopulateFiltersResponse, void>({
```
SamuelTilly, line 66: We can invalidate the cached filter result so that filter gets called again if the user change unit.

Add `providesTags: ['SickLeavesFilter'],` to `getPopulatedFilters` and `invalidatesTags: ['SickLeavesFilter'],` to `changeUnit`. `tagTypes` also needs to be updated with the new tag.

apps/rehabstod/src/components/Table/TableHeaderCell.tsx
```diff
@@ -18,6 +18,7 @@ export function TableHeaderCell<T extends string>({
 }) {
   return (
     <th
+      tabIndex={0}
       onClick={() => onColumnSort(column)}
```
SamuelTilly, line 22: We could add `onKeyDown` event handler here to make enter and space key trigger the `onColumnSort` function

SamuelTilly (approved):
Looks good 👍 could use some minor tweaks

PR #812, opened 2023-04-05
INTYGFV-15502: Added logic that sets previousPatientId if reserveId i…

ChrilleSigerhed:
…s true and previousPatientId exists, otherwise the patientId is set.


mww-sopra-steria (approved):
Nice!

PR #818, opened 2023-04-11
Add scroll to table

SamuelTilly:
Adding scroll to table and added custom tooltip component that works inside overflow hidden/scroll areas.


ChrilleSigerhed (approved):
Looks great, well done! Have we been able to verify this scroll behaviour?

PR #840, opened 2023-04-19
INTYGFV-15521: Fixed so that selected care unit is selected correctly…

ChrilleSigerhed:
… when user selects care unit to log in on.


mhornfeldt (approved):

UlrikaSopra (approved):
Nice!

PR #846, opened 2023-04-21
INTYGFV-15510: Fix issues with icons, input fields and SelectMultiple

SamuelTilly:
This aims to fix

1. Checkbox icons are not re-rendered correctly inside `SelectMultiple` dropdown
2. `SelectMultiple` does not respect viewbox boundaries.
3. Checkboxes inside `SelectMultiple` should gain focus when opened.
4. Wrong font size for input fields
5. Update `ids-core` to 2.6.0


apps/rehabstod/src/components/Form/NumberInput.tsx
```diff
@@ -1,49 +1,36 @@
...
   max?: number | string
   min?: number | string
-  className?: string
-  classNameInput?: string
-  novalidation?: boolean
+  error?: boolean
 }) {
+  const id = useId()
   return (
-    <IDSInput className={`${isRange && 'flex items-baseline gap-3'} ${className}`} novalidation={novalidation}>
```
SamuelTilly, line 33: `IDSInput` has currently multiple styling issues, border for error input is not rendered correctly. Full-width is not respected. Renamed "isRange" to "inline" to describe layout option.

apps/rehabstod/src/components/Form/NumberInput.tsx
```diff
@@ -1,49 +1,36 @@
...
   onBlur?: () => void
   description?: string
-  id: string
   value: number | string
-  isRange?: boolean
+  inline?: boolean
   max?: number | string
   min?: number | string
-  className?: string
-  classNameInput?: string
```
SamuelTilly, line 29: `className` on reusable components is problematic when refactoring later. It's difficult to know how a component is used and looks in different places when it's allowed to be changed with styling.

Some guideline
- Make sure a component always uses the full width and wrap with an element that constrains the component to fit within the layout.
- Add flags that describes the different states of a component, like with `error` here.
- Add types or styles to describe different styles, look at `DynamicLink` for example.

apps/rehabstod/src/components/Form/SelectMultiple.tsx
```diff
@@ -1,31 +1,104 @@
...
   if (!options || (options instanceof Array && options.length === 0)) {
     return null
   }
 
   return (
-    <>
+    <div>
       <label htmlFor={id}>{label}</label>
       {description && <TooltipIcon description={description} name="question" size="s" className="relative top-1 ml-2" />}
-      <IDSSelectMultiple placeholder={placeholder} className="text-sm">
```
mhornfeldt, line 26: We set the text to sm during a UX-meeting so that the lists wouldn't be to large.

SamuelTilly: This makes it different size from the NumberInput, looks really strange imo. I'll talk to UX and get a second opinion.

mhornfeldt: Otherwise maybe we can introduce a scroll later

apps/rehabstod/src/components/SettingsDialogContent/SettingsDialogContent.tsx
```diff
@@ -16,7 +17,17 @@ export function SettingsDialogContent({ preferences, onClose }: { preferences: U
     return null
   }
 
-  const isValueBetweenLimits = (max: number, min: number, value: string) => value && Number(value) <= max && Number(value) >= min
+  const isMaxAntalDagarMellanIntygValid = isValueBetweenLimits(
```
mhornfeldt, line 20: This has introduced a bug where the user can save empty inputs

SamuelTilly: Good catch, fixed

mhornfeldt (approved):

apps/rehabstod/src/components/Form/SelectMultiple.tsx
SamuelTilly: Talked to UX, it should be 16 for input fields, but it's ok for table cells to have 14px font size. Perhaps there was some confusion here. But we have the go-ahead for removing "text-sm" 😃

PR #847, opened 2023-04-21
INTYGFV-15548: Fixed bug with patient header that should li…

UlrikaSopra:
…e above the tablewith patientheader to overlap table


apps/rehabstod/src/components/Layout/LayoutHeader.tsx
```diff
@@ -14,7 +14,7 @@ export function LayoutHeader() {
   const ref = useRef<IDSDialogElement>(null)
 
   return (
-    <IDSHeader type="inera-admin" unresponsive>
+    <IDSHeader type="inera-admin" unresponsive className="z-50">
```
SamuelTilly, line 17: I think we can decrease this to z-40 and keep 50 for tooltips/popups.

apps/rehabstod/src/pages/Patient/components/PatientHeader.tsx
```diff
@@ -14,7 +14,7 @@ export function PatientHeader({ patient }: { patient: Patient }) {
   }
 
   return (
-    <div className="bg-secondary-95 sticky top-0">
+    <div className="bg-secondary-95 sticky top-0 z-40">
```
SamuelTilly, line 17: Decrease to 30. We can also make `TableHeaderCell` have `z-[11]` to be just above sticky rows at z-10.

SamuelTilly (changes requested):
We also need to change the z-index for `TooltipContent` so that it stays above everything else.

SamuelTilly (approved):

PR #856, opened 2023-04-25
INTYGFV-15558: Fixed expanding of settings menu.

ChrilleSigerhed:


mhornfeldt (approved):

mhornfeldt:
It would be a nicer solution if we didn't allow the SettingsDialog to control components of its parent (LayoutHeader), but I understand if it's not possible

PR #864, opened 2023-04-27
INTYGFV-15527: Added option to select unit and remove saved unit from SettingsDialog.

ChrilleSigerhed:


apps/rehabstod/src/components/Form/SelectCareUnits.tsx
```diff
@@ -0,0 +1,48 @@
...
+          careUnit.mottagningar.forEach((reception) => units.push(reception))
+        }
+      })
+    })
+    return units
+  }
+
+  if (!user) {
+    return null
+  }
```
SamuelTilly, line 26: We could simplify this a bit

```typescript
if (!user) {
  return null
}

const units: (Vardenhet | Mottagning)[] = user.vardgivare
  .map(({ vardenheter }) => vardenheter.map((careUnit) => [careUnit, ...(careUnit.mottagningar ?? [])]).flat())
  .flat()
```

apps/rehabstod/src/components/Form/SelectCareUnits.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+      numberOfUnits += 1
+      if (careUnit.mottagningar && careUnit.mottagningar.length > 0) {
+        careUnit.mottagningar.forEach(() => {
+          numberOfUnits += 1
+        })
+      }
+    })
+  })
+  return numberOfUnits
+}
```
SamuelTilly, line 32: This could use the same `getUnits` function from the components with some tweaks if we move it out into a utils folder.

apps/rehabstod/src/components/Form/SelectCareUnits.tsx
SamuelTilly, line 1: I think this component belongs in the `SettingsDialog` folder, "Form" can be reserved for simple reusable form components. This component fills a specific function for user units.

apps/rehabstod/src/components/Form/SelectCareUnits.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <IDSSelect>
+      {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
SamuelTilly, line 30: We can fix this rule to better work with IDS components. Please add this to `.eslintrc.cjs` rules.
```
'jsx-a11y/label-has-associated-control': [
  'error',
  {
    labelComponents: [],
    labelAttributes: [],
    controlComponents: [],
    assert: 'either',
    depth: 25,
  },
],
```

This copies the rules from airbnb but allows us to have label as sibling to the input https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react-a11y.js#L123-L129

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/label-has-associated-control.md

The component also needs an ID between select and label, so please add `htmlFor={id}` to the label and `id={id}` to the select element. You can generate an ID with the React function `const id = useId()`

SamuelTilly (changes requested):
Nice work! Some tweaks suggested

apps/rehabstod/src/components/Form/SelectCareUnits.tsx
ChrilleSigerhed: Great suggestion!

apps/rehabstod/src/components/Form/SelectCareUnits.tsx
ChrilleSigerhed: Alright, makes sense! Done!

apps/rehabstod/src/components/Form/SelectCareUnits.tsx
ChrilleSigerhed: Thanks for clarifying how to fix this issue. Worked great!

apps/rehabstod/src/components/SettingsDialog/SelectCareUnits.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+const preferences = fakeUserPreferences()
+let onChange: (value: string) => void
+let user: User
+beforeEach(() => {
+  onChange = vi.fn()
+  user = fakeUser()
+})
+
+it('should display Ingen forvald enhet option', async () => {
+  server.use(rest.get('/api/user', (_, res, ctx) => res(ctx.status(200), ctx.json(user))))
```
SamuelTilly, line 21: Looks like this could be moved to `beforeEach`

apps/rehabstod/src/components/SettingsDialog/SelectCareUnits.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  renderWithRouter(<SelectCareUnits onChange={onChange} preferences={userPreferences} />)
+
+  expect(await screen.findByDisplayValue(unit.namn)).toBeInTheDocument()
+})
+
+it('should call onChange if option is selected', async () => {
+  const unit = user.vardgivare[0].vardenheter.filter((units) => units.id !== undefined)[0]
+
+  server.use(rest.get('/api/user', (_, res, ctx) => res(ctx.status(200), ctx.json(user))))
+  renderWithRouter(<SelectCareUnits onChange={onChange} preferences={preferences} />)
```
SamuelTilly, line 48: Looks like `onChange` is only used here, we could define and move it to this test.

apps/rehabstod/src/components/SettingsDialog/SelectCareUnits.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <IDSSelect>
+      <label className="mb-0" htmlFor={id}>
+        Förvald enhet
+      </label>
+      <select
+        id={id}
+        name="options"
+        className="text-neutral-20 my-3 box-border w-full truncate rounded border py-3 px-5 text-left"
```
SamuelTilly, line 22: I believe we can remove `my-3` to get rid of that extra margin we noticed in the demo. `IDSSelect` adds the same margin so it probably became doubled.

SamuelTilly (approved):
Looking good! Please fix the margin, but the suggestions to test is optional ;D

apps/rehabstod/src/components/SettingsDialog/SelectCareUnits.tsx
```diff
@@ -0,0 +1,34 @@
...
+export function SelectCareUnits({ onChange, preferences }: { onChange: (value: string) => void; preferences: UserPreferences }) {
+  const { isLoading, data: user } = useGetUserQuery()
+  const id = useId()
+  if (!user || isLoading) {
+    return null
+  }
+
+  return (
+    <IDSSelect>
+      <label htmlFor={id}>Förvald enhet</label>
```
SamuelTilly, line 16: Talked to UX, this should be changed to "Välj enhet"

SamuelTilly (changes requested):

SamuelTilly (approved):

PR #866, opened 2023-04-27
INTYGFV-15556: Input should only allow numbers not decimal signs

mhornfeldt:


UlrikaSopra (approved):
But if I copy and paste it still works with ,

PR #881, opened 2023-05-04
INTYGFV-15583: Updated sick leave length filter to support intervals

mhornfeldt:


ChrilleSigerhed (dismissed):
Looks great, one small thing! After clicking on one of the listed sick leaves, and navigated to the patient view, and then when returning back from that view the state seems to be resetted.

I think we should fix this so that it works the same as for Diagnosis & Doctor-filter
image

SamuelTilly (changes requested):

SamuelTilly (approved):

PR #882, opened 2023-05-04
INTYGFV-15572: Updated logic related to patient page with new encrypt…

ChrilleSigerhed:
…ed patientId.

### DIFF
--- a/apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.test.tsx
+++ b/apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.test.tsx
@@ -67,7 +67,7 @@ describe('Navigate', () => {
                 <TableBodyRows
                   isLoading={false}
                   showPersonalInformation
-                  sickLeaves={[fakerFromSchema(sickLeaveInfoSchema)({ patient: { id: 'aperiam' } })]}
+                  sickLeaves={[fakerFromSchema(sickLeaveInfoSchema)({ uid: 'aperiam' })]}
                   unitId="Alfa Vårdenhet"
                   isDoctor={false}
                 />
@@ -75,7 +75,7 @@ describe('Navigate', () => {
             </Table>
           }
         />
-        <Route path={`/pagaende-sjukfall/${btoa('aperiam')}`} element={<p>Patient Route</p>} />
+        <Route path="/pagaende-sjukfall/aperiam" element={<p>Patient Route</p>} />
       </Routes>
     )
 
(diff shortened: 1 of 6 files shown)


apps/rehabstod/src/pages/Patient/Patient.tsx
```diff
@@ -10,7 +10,7 @@ import { PatientOverview } from './components/patientOverview/PatientOverview'
 export function Patient() {
   const { patientId } = useParams()
```
ChrilleSigerhed, line 11: Should this variable be renamed to something like encryptedPatientId, to better mirror the value? Im open for suggestion.

mhornfeldt: I think encryptedPatientId makes the code more understandable

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.test.tsx
```diff
@@ -67,15 +67,15 @@ describe('Navigate', () => {
...
+                  sickLeaves={[fakerFromSchema(sickLeaveInfoSchema)({ uid: 'aperiam' })]}
                   unitId="Alfa Vårdenhet"
                   isDoctor={false}
                 />
               </tbody>
             </Table>
           }
         />
-        <Route path={`/pagaende-sjukfall/${btoa('aperiam')}`} element={<p>Patient Route</p>} />
+        <Route path="/pagaende-sjukfall/aperiam" element={<p>Patient Route</p>} />
```
mhornfeldt, line 78: Is this correct?

mhornfeldt (approved):

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.test.tsx
ChrilleSigerhed: I believe so. From what I can tell we previously set 'aperiam' to patient.id. This has been updated to uid instead. Since the uid is already encrypted, there is no need to Base64.encode the string?

apps/rehabstod/src/pages/Patient/Patient.tsx
ChrilleSigerhed: updateded!

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.test.tsx
mhornfeldt: Ah sorry I didn't see it was a test. I thought you were hard coding a url.

PR #884, opened 2023-05-04
INTYGFV-15556, INTYGFV-15577: SettingsDialog issues

SamuelTilly:
* INTYGFV-15577: Fix wrong background color for Select element
* INTYGFV-15556: Fix issue with entering leading zeros
* Fix issue with `IDSHeaderAvatar` not closing when dialog is shown.
* Merge `SettingsDialog` and `SettingsDialogContent` to avoid having to propagate events.
* Introduce settings slice to hold `SettingsDialog` state
* Replace validation function with `zod` validation
* Added cleanup for `changedVisibility` event
* Added missing `IDSDialogActions` from ids-core


apps/rehabstod/src/components/SettingsDialog/DaysFinishedSickLeave.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <div className="[&:not(:last-child)]:mb-5">
+      <h2 className="ids-heading-4">Visa nyligen avslutade sjukfall</h2>
+      <p className="pb-4">
+        Välj maximalt antal dagar som får ha passerat efter ett sjukfalls slutdatum för att sjukfallet ska visas upp i sjukfallstabellen.
+        Med denna funktion kan du bevaka de sjukfall som är nyligen avslutade.
+      </p>
+      <div className="w-80">
+        <FormattedNumberInput
+          label="Max antal dagar sedan avslut  (0-14 dagar)"
```
mhornfeldt, line 21: I think there's some extra spacing here

apps/rehabstod/src/schemas/userSchema.ts
```diff
@@ -2,6 +2,15 @@ import { z } from 'zod'
 import { vardenhetSchema } from './vardenhetSchema'
 import { vardgivareSchema } from './vardgivareSchema'
 
+export const DAYS_BETWEEN_SICK_LEAVES = {
+  MIN: 0,
```
mhornfeldt, line 6: There's  a bug when I try to enter 0 it resets the value to the last saved value for both inputs. I can see here that the min limit is correct. Maybe it reads 0 as empty string and therefore resets it?

apps/rehabstod/src/components/SettingsDialog/SelectCareUnits.tsx
```diff
@@ -2,34 +2,43 @@ import { IDSSelect } from '@frontend/ids-react-ts'
...
+      <h2 className="ids-heading-4">Förvald enhet</h2>
+      <p>
+        Du kan välja en enhet som du automatiskt loggas in på när Rehabstöd startas. Välj &quot;Ingen förvald enhet&quot; i listan för att
+        rensa ditt val.
+      </p>
+      <div className="w-80">
+        <IDSSelect>
+          <label htmlFor={id}>Välj enhet</label>
+          <div>
+            <Select
```
mhornfeldt, line 25: The spacing between the label and the input is larger than for the number inputs

apps/rehabstod/src/store/slices/settings.slice.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-param-reassign */
+import { createSlice, PayloadAction } from '@reduxjs/toolkit'
```
mhornfeldt, line 2: The local state has introduced a bug where I can enter values and click "cancel" but the values still show up in the modal if I open it again. It does however not save it as the user's settings. We need to reset the state to the saved values if we close the modal.

apps/rehabstod/src/store/slices/settings.slice.ts
```diff
@@ -0,0 +1,35 @@
...
+import { UserPreferences } from '../../schemas'
```
mhornfeldt, line 3: The dialog can't be closed with esc anymore

mhornfeldt (changes requested):

apps/rehabstod/src/components/SettingsDialog/DaysFinishedSickLeave.tsx
SamuelTilly: Sneaky, should be fixed now :)

apps/rehabstod/src/schemas/userSchema.ts
SamuelTilly: Good catch, changed `convertValue` to look for `!Number.isNaN(val)` instead of `Boolean(val)` solved the issue.

apps/rehabstod/src/components/SettingsDialog/SelectCareUnits.tsx
SamuelTilly: Fixed 👍

apps/rehabstod/src/store/slices/settings.slice.ts
SamuelTilly: Good catch! added a reset dispatch, should work as expected now.

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableInfo.tsx
```diff
@@ -27,22 +35,22 @@ export function TableInfo({
...
-        </p>
-        <p className="mx-2">|</p>
-        <p>
-          <span className="font-bold">{daysBetweenCertificates} dagar</span> mellan intyg
-        </p>
-        <p className="mx-2">|</p>
-        <p>
+        </span>
+        <TableInfoDivider />
+        <span className="font-bold">{daysBetweenCertificates} dagar</span> mellan intyg
```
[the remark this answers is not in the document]
SamuelTilly: Fixed 👍

apps/rehabstod/src/store/slices/settings.slice.ts
SamuelTilly: Adding focus to first available input when shown seems to have solved the issue.

SamuelTilly:
> Seems like the state does not reset to the original values in the SettingsDialog if the user chooses to close the modal without saving the new changes.
> 
> Im guessing the saved values in preferences should be displayed instead?

Wops, my bad. dispatched a save event instead of resetting settings state. Should be fixed now.

ChrilleSigerhed (approved):
Awsome! Now the SettingsDialog works as expected!

mhornfeldt (approved):

PR #898, opened 2023-05-09
INTYGFV-15366: Printable sick leave table

SamuelTilly:


mhornfeldt (approved):
Nice! Should some tests be added?

SamuelTilly:
> Nice! Should some tests be added?

Good call, I'll add a couple of tests

PR #926, opened 2023-05-29
Feature/move to pnpm

SamuelTilly:


ChrilleSigerhed (approved):
Excellent work, im glad that this solved our problems regarding dependency errors.

Maybe we should schedule a session where we go over the changes that this will require from the developers and how it may affect our workflow before merging this?

SamuelTilly:
> Maybe we should schedule a session where we go over the changes that this will require from the developers and how it may affect our workflow before merging this?

Not sure that I have enough information for a whole meeting. Things will work mostly the same, replace yarn with pnpm and replace "workplace" with "--filter".

PR #930, opened 2023-05-30
INTYGFV-15661: Added OccupationFilter to Filters.

ChrilleSigerhed:


apps/rehabstod/src/pages/CurrentSickLeaves/utils/getOccupationPlaceholder.ts
```diff
@@ -0,0 +1,14 @@
+import { OccupationType } from '../../../schemas/sickLeaveSchema'
+
+export const getOccupationPlaceholder = (selected: string[], occupationTypes: OccupationType[]) => {
```
mhornfeldt, line 3: This is the same as the rekoStatus placeholder right? Maybe we could merge these two since the type objects look the same we could send in types: OccupationType[] | RekoStatusType[]

mhornfeldt: Another option is that they take the same object for example TextIdFIlterType or something similar, as long as the id and name inside the object have the same name it will work against the backend

mhornfeldt (changes requested):

apps/rehabstod/src/pages/CurrentSickLeaves/utils/getOccupationPlaceholder.ts
ChrilleSigerhed: From what I could tell there is a difference in that when no options are selected, occuaptionPlacerholder should be "Alla valda". Maybe this is also true for Reko? in that case I totally agree!

apps/rehabstod/src/pages/CurrentSickLeaves/components/filter/OccupationFilter.tsx
```diff
@@ -0,0 +1,48 @@
+import { Checkbox } from '../../../../components/Form/Checkbox'
```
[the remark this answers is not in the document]
ChrilleSigerhed: Good suggestion, the only difference is the placerholder that will affect the tests. I will look into it!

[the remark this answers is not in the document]
mhornfeldt: I think the placeholder should be the same but if they're not we don't need to test the placeholder in the filter class since it has its own tests

PR #934, opened 2023-05-31
INTYGFV-15659: Added srs risk signal to sick leave table

mhornfeldt:


SamuelTilly (changes requested):
Looks good!

Remove or fix the unused test and add the new column to `PrintTable` component

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableHeaderRow.test.tsx
```diff
@@ -35,17 +36,17 @@ it('Should render all but doctor column if user is doctor', async () => {
   expect(screen.queryByRole('columnheader', { name: 'Läkare' })).not.toBeInTheDocument()
 })
 
-it.each(Object.values(SickLeaveColumn))('Should render and hide %s column', async (column) => {
-  renderWithRouter(
```
SamuelTilly, line 39: Could we keep this tests and just add `server.use(rest.get('/api/sickleaves/filters', (_, res, ctx) => res(ctx.status(200), ctx.json({ srsActivated: true }))))` before `renderWithRouter` is called?

SamuelTilly (approved):
🥇

PR #940, opened 2023-06-02
INTYGFV-15656: Added unanswered communication to sick leave table

mhornfeldt:


SamuelTilly (changes requested):
We have most of this functionality in Patient view already, could we reuse the `getQAStatusFormat` function used there?

SamuelTilly (approved):

PR #946, opened 2023-06-07
Remove or fix tests containing timeouts

SamuelTilly:


packages/webcert/src/feature/certificate/Modals/CreateCertificateFromCandidateWithMessageModal.test.tsx
```diff
@@ -98,15 +98,14 @@ describe('Create certificate from candidate modal', () => {
     expect(screen.getByText('Avbryt')).toBeInTheDocument()
   })
 
-  it('When show is clicked, new message and show button removed, but cancel button remaining', () => {
+  it.skip('When show is clicked, new message and show button removed, but cancel button remaining', async () => {
```
SamuelTilly, line 101: This test is not working, and probably never has. Due to using `setTimeout` it has been terminated before assertions inside the timeout has been executed.

PR #948, opened 2023-06-07
INTYGFV-15676: Removed caching from query getSickLeavePatient thats u…

ChrilleSigerhed:
…sed in Patient.tsx.

### DIFF
--- a/apps/rehabstod/src/store/api.ts
+++ b/apps/rehabstod/src/store/api.ts
@@ -123,11 +123,12 @@ export const api = createApi({
       }),
       providesTags: ['SickLeaveSummary'],
     }),
-    getSickLeavePatient: builder.query<Patient, { encryptedPatientId: string }>({
-      query: ({ encryptedPatientId }) => ({
+    getSickLeavePatient: builder.query<Patient, { encryptedPatientId: string | undefined }>({
+      keepUnusedDataFor: 0,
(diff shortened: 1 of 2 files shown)


ChrilleSigerhed:
Since we are not able to invalidateTags on queries, I decided to remove the caching on the getSickLeavePatient by setting "keepUnusedData" to 0.

I tried the recommended ```refetchOnMountOrArgChange ``` but got a weird update after the component already was rendererd.

```javascript
const { data: patient, error } = useGetSickLeavePatientQuery({ encryptedPatientId }, { refetchOnMountOrArgChange: true })
```

apps/rehabstod/src/store/api.ts
```diff
@@ -123,11 +123,12 @@ export const api = createApi({
       }),
       providesTags: ['SickLeaveSummary'],
     }),
-    getSickLeavePatient: builder.query<Patient, { encryptedPatientId: string }>({
-      query: ({ encryptedPatientId }) => ({
+    getSickLeavePatient: builder.query<Patient, { encryptedPatientId: string | undefined }>({
```
SamuelTilly, line 126: We should probably not allow fetching with undefined `encryptedPatientId`. Please revert this change.

SamuelTilly (changes requested):

SamuelTilly (approved):

PR #958, opened 2023-06-09
Run prettier independent from linter

SamuelTilly:
It's not recommended to use [eslint-plugin-prettier](https://prettier.io/docs/en/integrating-with-linters.html). Instead, it's recommended to use prettier --check to verify formatting is correct. This will simplify prettier integration a bit and hopefully fix the issues we are facing with ESLint since changing to pnpm.

* Update prettier to latest version
* Update ESLint to latest version
* Add step to GitHub actions for verifying formatting
* Add prettier to pre-commit hook
* Fix pre-commit issue on Windows
* Patch current version of ids-core (issue with loading unlisted dependency lit-html)
* Remove eslint-plugin-prettier


ChrilleSigerhed (approved):
Awsome!! Great work, this will improve our development experience alot!

Great that u formatted all the files aswell!

PR #965, opened 2023-06-12
INTYGFV-15594: Added tooltips to pie chart

mhornfeldt:


ChrilleSigerhed (approved):
Works like a charm! Great work!

PR #969, opened 2023-06-12
INTYGFV-15625: Add ErrorBoundary component for client errors

SamuelTilly:


apps/rehabstod/src/components/error/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { current: errorId } = useRef(uuidv4())
+  const message = errorMessage(error)
+  const stackTrace = errorStacktrace(error)
+
+  useEffect(() => {
+    dispatch(
+      api.endpoints.logError.initiate({
+        errorData: {
+          errorId,
+          errorCode: 'No errorCode',
```
ChrilleSigerhed, line 40: I suggest we add CLIENT_ERROR as errorCode.

ChrilleSigerhed (approved):
I like the refactorization you did, and also how you restructured the files in the project. Makes sense to structure it like this!

Cool that you fetched errorId throught useRef, looks cleaner!

I noticed when I tested out the new functionality that we seem to always log the error twice, is that something we can fix or it is something we have to live with?

Left a suggestion regarding errorCode, other than that, great work!!!

SamuelTilly:
> I noticed when I tested out the new functionality that we seem to always log the error twice, is that something we can fix or it is something we have to live with?

Good catch, I fixed it so that it only log the error once.

PR #974, opened 2023-06-15
INTYGFV-15388: Added LU table and more dynamic table logic

mhornfeldt:
…U certificates


apps/rehabstod/src/components/Table/tableBody/TableRow.tsx
```diff
@@ -0,0 +1,35 @@
+import { ReactNode } from 'react'
+
+export function TableRow({
+  italic,
+  onClick,
+  id,
+  children,
+}: {
+  italic: boolean
+  onClick: (key: string) => void
```
SamuelTilly, line 10: onClick sounds like we would be expecting a mouse event. Perhaps rename to onSelect, onNavigate or something 🤷‍♂️

apps/rehabstod/src/components/error/ErrorAlert/ErrorAlert.tsx
```diff
@@ -10,13 +10,15 @@ export function ErrorAlert({
...
   dynamicLink,
+  hideErrorId,
   error,
 }: {
   heading: string
   errorType: 'info' | 'attention' | 'success' | 'error'
   text: string
   dynamicLink: boolean
-  error: (FetchBaseQueryError & { id?: string }) | (SerializedError & { id?: string })
+  hideErrorId?: boolean
```
SamuelTilly, line 20: Is this used? Could not find an instance where we want to hide error id.

apps/rehabstod/src/components/error/GetTableContentError/GetTableContentError.tsx
SamuelTilly, line 1: We usually don't use verbs like `Get` in component name. Perhaps change to  `TableContentError`

apps/rehabstod/src/components/error/GetTableContentError/GetTableContentError.tsx
```diff
@@ -0,0 +1,22 @@
...
+export function GetTableContentError({
+  tableName,
+  error,
+}: {
+  tableName: string
+  error: (FetchBaseQueryError & { id?: string }) | (SerializedError & { id?: string })
+}) {
+  return (
+    <ErrorAlert
+      heading={getTableErrorTitle(tableName)}
```
SamuelTilly, line 15: This returns a text that is not reused anywhere else. Why not include it in this component?

apps/rehabstod/src/components/error/GetTableContentError/GetTableContentError.tsx
```diff
@@ -0,0 +1,22 @@
...
+      errorType="error"
+      text={getTableErrorDescription(tableName)}
```
SamuelTilly, line 17: Same with this text

apps/rehabstod/src/components/error/UnansweredCommunicationError/UnansweredCommunicationError.tsx
SamuelTilly, line 1: This could be moved to `ErrorAlert/UnansweredCommunicationError` and add `Alert` as suffix to better present what this component is.

apps/rehabstod/src/components/Table/TableLayout.tsx
```diff
@@ -0,0 +1,69 @@
...
+  useEffect(() => {
+    if (!isUserLoading && !user) {
+      navigate('/')
+    }
+  }, [user, isUserLoading, navigate])
+
+  return (
+    <div className="ids-content m-auto max-w-7xl py-10 px-2.5">
+      {heading}
+      <h3 className="ids-heading-4 hidden print:block">Valda filter</h3>
```
SamuelTilly, line 46: Could this be moved to the `TableFilter` component?

apps/rehabstod/src/pages/LUCertificates/LUCertificates.tsx
```diff
@@ -0,0 +1,110 @@
...
+          listLength={luCertificatesInfo ? luCertificatesInfo.certificates.length : 0}
+          totalNumber={luCertificatesInfo ? luCertificatesInfo.certificates.length : 0} // TODO: this should come from backend
+          showPersonalInformation={showPersonalInformation} // TODO: should be shared with sickleaves
+          onShowPersonalInformationChange={(checked) => dispatch(updateShowPersonalInformation(checked))}
+        />
+      }
+      modifyTableColumns={<ModifyLUCertificatesTableColumns />}
+      error={error}
+      unansweredCommunicationError={!!luCertificatesInfo?.questionAndAnswersError}
+    >
```
SamuelTilly, line 74: For a layout component I would rather see it only accepting components, user, error, printable should be handled by the component's responsible for those things. Instead of error I would expect for example `fetchError={<GetTableContentError tableName={TABLE_NAME} error={error} />}`

apps/rehabstod/src/components/Table/TableLayout.tsx
```diff
@@ -0,0 +1,69 @@
...
+  printable: boolean
+  children: ReactNode
+}) {
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    if (!isUserLoading && !user) {
+      navigate('/')
+    }
+  }, [user, isUserLoading, navigate])
```
SamuelTilly, line 41: This could be moved to a shared hook instead feels out of place for a general purpose component.

apps/rehabstod/src/components/Table/TableLayout.tsx
```diff
@@ -0,0 +1,69 @@
...
+            <div className="w-full">{tableInfo}</div>
+            <div className="mb-5 flex items-end gap-3 print:hidden">
+              <div className="w-96">{modifyTableColumns}</div>
+              {printable && (
+                <IDSButton onClick={() => window.print()} className="mb-3 whitespace-nowrap">
+                  Skriv ut
+                </IDSButton>
+              )}
+            </div>
+          </div>
```
SamuelTilly, line 63: Perhaps this could be made into a component.

apps/rehabstod/src/pages/LUCertificates/utils/luCertificatesTableValueFormatter.tsx
```diff
@@ -0,0 +1,48 @@
...
+        </TableCell>
+      )
+    case LUCertificatesColumn.Signeringsdatum:
+      return <TableCell>{(getLUCertificatesTableValue(column, data) as string).split('T')[0]}</TableCell>
+    case LUCertificatesColumn.Ärenden:
+      return <TableCell>{getUnansweredCommunicationFormat(data.unAnsweredComplement, data.unAnsweredOther)}</TableCell>
+    default:
+      return <TableCell>{getLUCertificatesTableValue(column, data)}</TableCell>
+  }
+}
```
SamuelTilly, line 48: This is returning `JSX.Element`.  It should be moved and made into a component.

apps/rehabstod/src/components/Table/heading/TableHeadingForUnit.tsx
```diff
@@ -0,0 +1,15 @@
+import { User, UserUrval } from '../../../schemas'
+import { TableHeading } from './TableHeading'
+
+export function TableHeadingForUnit({ user, tableName, suffix }: { user?: User; tableName: string; suffix?: string }) {
+  const isDoctor = user?.urval === UserUrval.ISSUED_BY_ME // TODO: refactor to utils
```
SamuelTilly, line 5: Should this use `src/utils/isUserDoctor.ts`

SamuelTilly (changes requested):
Nice work!

apps/rehabstod/src/components/error/ErrorAlert/ErrorAlert.tsx
mhornfeldt: Yes for https://inera.atlassian.net/browse/INTYGFV-15725 but I can omit the error instead.

apps/rehabstod/src/components/Table/tableBody/TableBody.tsx
```diff
@@ -0,0 +1,42 @@
...
+}) {
+  const { sortTableList } = useTableContext()
+
+  return (
+    <>
+      {sortTableList(content, tableValueExtractor).map(
+        (item) =>
+          columns.length > 0 && (
+            <TableRow
+              key={item[keyIndex] as string}
```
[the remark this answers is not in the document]
mhornfeldt: T is either a Sjukfall or a Läkarutlåtande. Since the objects are not always including string values how should I type this? Can I type that the keyIndex matches a value that is always a string instead?

apps/rehabstod/src/pages/LUCertificates/LUCertificates.tsx
mhornfeldt: What are the benefits of the component only accepting components?

apps/rehabstod/src/pages/LUCertificates/utils/luCertificatesTableValueFormatter.tsx
mhornfeldt: I don't agree with this I think it will make the TableBody less understandable

apps/rehabstod/src/pages/LUCertificates/LUCertificates.tsx
SamuelTilly: It's to help avoid mixing logic with layout

mhornfeldt: Creating the hook removed some attributes but I think the ones that are there now should stay since they toggle components which are common for all tables.

apps/rehabstod/src/components/Table/tableBody/TableBody.tsx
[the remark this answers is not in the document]
SamuelTilly: This is the true challenge of making things dynamic, not exactly sure how this should be typed. Currently, this is hiding an issue where we are using `patient` object as key and id where we should be using `patient.id`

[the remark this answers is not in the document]
mhornfeldt: That problem is only there because we haven't had time to decorate LU with encryptedPatientId. We will always use encryptedPatientId later which is always a string and not inside an object

apps/rehabstod/src/pages/LUCertificates/utils/luCertificatesTableValueFormatter.tsx
SamuelTilly: With this solution, the array of element is not properly assigned key and React can't handle re-rendering properly.

mhornfeldt: If I change it to a component how can I write it in a way that works for several tables? Also, I don't get any errors in the console that there should be problems with the re-rendering of the table

apps/rehabstod/src/components/Table/TableLayout.tsx
```diff
@@ -0,0 +1,53 @@
...
+  filters: ReactNode
+  tableInfo: ReactNode
+  modifyTableColumns: ReactNode
+  tableContentError?: (FetchBaseQueryError & { id?: string }) | (SerializedError & { id?: string })
+  unansweredCommunicationError?: boolean
+  tableName: string
+  printable: boolean
+  children: ReactNode
+}) {
+  useNavigateToStartPage()
```
SamuelTilly, line 30: Nice hook 😄 but it should be moved to `pages/LUCertificates`

apps/rehabstod/src/pages/LUCertificates/utils/luCertificatesTableValueFormatter.tsx
mhornfeldt: Added a potential solution for this

apps/rehabstod/src/components/Table/TableLayout.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <div className="ids-content m-auto max-w-7xl py-10 px-2.5">
+      {heading}
+      {filters}
+      {tableContentError ? (
+        <TableContentAlert tableName={tableName} error={tableContentError} />
+      ) : (
```
SamuelTilly, line 38: We can make the top part into a component, this is also used by the patient page, but the layout is a bit different.
```tsx
export function PageContainer({ children }: { children: ReactNode }) {
  return <div className="ids-content m-auto max-w-7xl py-10 px-2.5">{children}</div>
}
```

And we can move out heading, filters and children. `LUCertificate` would look something like:

```tsx
 <PageContainer>
  <TableHeadingForUnit ... />
  <LUCertificatesFilters ... />
  {error && <TableContentAlert ... /> }
  {luCertificatesInfo?.questionAndAnswersError && (
    <div className="pb-10">
      <UnansweredCommunicationAlert />
    </div>
  )}
  <TableLayout printable tableInfo={...} modifyTableColumns={...} />
  {!error && (
        <Table ... />}
</PageContainer>
```

This would make it so `TableLayout` only holds `tableInfo` and `modifyTableColumns`, the name is no longer great, but this makes it so that we are a bit more flexible.

apps/rehabstod/src/components/Table/tableBody/TableBody.tsx
```diff
@@ -0,0 +1,45 @@
...
+  tableValueExtractor: <S extends T>(column: string, data: S) => unknown
+  tableIdExtractor: <S extends T>(data: S) => string
+  onTableRowClick: (key: string) => void
+  columns: TableColumn[]
+  TableCellResolverComponent: React.Component
+}) {
+  const { sortTableList } = useTableContext()
+
+  return (
+    <>
```
SamuelTilly, line 26: Missing `<tbody>`

apps/rehabstod/src/components/Table/TableLayout.tsx
mhornfeldt: If it's only used in tables then I think the TableLayout is a good place and it's not necessary to move it into a separate component. If we feel like we want to move it out to a separate container when refactoring the patient page we can do that then. I will leave the TableLayout as it is now since I don't think that moving out the heading etc. presents more benefits.

mhornfeldt: I'm gonna refactor CurrentSickleaves when this is merged so we can probably find even more improvements/changes then

SamuelTilly: > If it's only used in tables then I think the TableLayout is a good place

It's not, that is why I suggested the change
image

This `div` could be considered general page layout, it would be inside Layout if it wasn't for Patient having a header that needs to be rendered above this element. Therefore, it's not really part of a Table construct and should be moved. This change in turn renders `TableLayout` a bit redundant except for holding the layout for `TableInfo` and `ModifyTableColumns`

apps/rehabstod/src/components/Table/tableBody/TableBody.tsx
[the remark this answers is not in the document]
SamuelTilly: Looks like that introduced some new typescript issues.

I don't think the proposed solution brings enough value when they could just be rendered as components in the first place. If we break apart the differences and compare them

The proposed solution:
```tsx
export function LUCertificateTableBody({
  luCertificatesInfo,
  visibleColumns,
  onTableRowClick,
}: {
  luCertificatesInfo?: LUCertificatesInfo
  visibleColumns: TableColumn[]
  onTableRowClick: (key: string) => void
}) {
  return (
    <TableBody
      content={luCertificatesInfo?.certificates ?? []}
      tableValueExtractor={getLUCertificatesTableValue}
      tableIdExtractor={(data: LUCertificate) => data.patient.id}
      columns={visibleColumns}
      onTableRowClick={onTableRowClick}
      TableCellResolverComponent={LUCertificatesTableCellResolver}
    />
  )
}
```

The current solution:
```tsx
export function LUCertificateTableBody({
  certificates,
  columns,
  onTableRowClick,
}: {
  certificates: LUCertificate[]
  columns: TableColumn[]
  onTableRowClick: (key: string) => void
}) {
  const { sortTableList } = useTableContext()

  return (
    <tbody>
      {sortTableList(certificates, getLUCertificatesTableValue).map(
        (item) =>
          columns.length > 0 && (
            <TableRow focusable key={item.patient.id} id={item.patient.id} onNavigate={onTableRowClick}>
              {columns.map(({ name }) => (
                <LUCertificatesTableCellResolver key={name} column={name} data={item} />
              ))}
            </TableRow>
          )
      )}
    </tbody>
  )
}
```

Re-using one hook and a loop for the different tables is not that bad.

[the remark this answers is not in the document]
SamuelTilly: But print is handled by `PrintTableBody`, it doesn't get solved by this abstraction. This adds a callback pattern that we currently aren't using for very little gain. Similar to `TableBody`, `PrintTableBody` could be written like regular react code using components:

```tsx
return (
  <PrintBlock>
    {sortTableList(content, getLUCertificatesTableValue).map((item) => (
      <PrintTableRow>
        {columns.length > 0 &&
          columns.map(({ name }) => (
            <PrintTableCell key={name} name={name}>
              <LUCertificatesTableCellResolver key={name} column={name} data={item} />
            </PrintTableCell>
          ))}
      </PrintTableRow>
    ))}
  </PrintBlock>
)
```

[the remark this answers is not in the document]
mhornfeldt: I don't think it's little gain to add more generic components. When we create duplicated components with copied code we increase the risk of having to do bug fixes in several places, we open up to the risk that the components start to drift apart, and we enable a behavior of writing type (like SickLeave, Patient or LU) specific code in components that should actually be generic and not include certificate specific logic. We can see this example in the code in several places for the SickLeave-table and the Patient-table.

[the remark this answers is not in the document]
SamuelTilly: But `TableBody` and `PrintTableBody` holds essentially no logic at all, what are we trying to generalize with these component?

Just noticed he `columns.length > 0` check, it can be removed as well.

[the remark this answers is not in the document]
mhornfeldt: The points I mention above is not only affected by the amount of code that is saved by generalizing components. Like I wrote before it's also about not having individual components which do the exact same thing which later might drift apart and that we enable behavior of writing certificate specific logic in component which should not have this logic. The problem of where the logic is placed is something that the React part of Rehabstöd doesn't have a good division of.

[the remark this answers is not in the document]
SamuelTilly: The risk of components drifting apart is not limited by having these components

SamuelTilly (approved):

PR #975, opened 2023-06-15
INTYGFV-15653: Updated key valaue for PatientTableBoy since duplicade…

ChrilleSigerhed:
…d keys caused sorting failure.


mhornfeldt (approved):
Nice!

PR #977, opened 2023-06-19
INTYGFV-15644: Ärendestatus filter not persisting

SamuelTilly:


ChrilleSigerhed (approved):
Great, now the component works as expected!

Consider adding a test to verify that the value from filter-state in SelectFilter is used before merging!

PR #982, opened 2023-06-21
INTYGFV-15693: Changed so that we use the backend logic to determine …

ChrilleSigerhed:
…if certificate should be able to open in WC.


mhornfeldt (approved):
Nice!!

PR #983, opened 2023-06-22
INTYGFV-15727: SRS fix

mhornfeldt:


ChrilleSigerhed (approved):
Nice! If possible, add a test to verify this logic before merging!

mhornfeldt:
> Nice! If possible, add a test to verify this logic before merging!

It's been very hard to test these graphs because of the library sadly so I'll skip it for now. But in the future if we find a good way for sure!

PR #985, opened 2023-06-22
INTYGFV-15357: Set REKO-status in patient view

mhornfeldt:
I've asked requirements if REKO-status should also work if no current sick leaves but only previous. If that's the case I'll update the solution.

### DIFF
--- a/apps/rehabstod/src/components/SelectRekoStatus/SelectRekoStatus.tsx
+++ b/apps/rehabstod/src/components/SelectRekoStatus/SelectRekoStatus.tsx
@@ -1,0 +1,43 @@
+import { useState } from 'react'
+import { useSetRekoStatusMutation } from '../../store/api'
+import { RekoStatus, RekoStatusType } from '../../schemas/sickLeaveSchema'
+import { SelectFilter } from '../../pages/CurrentSickLeaves/components/filter/SelectFilter'
+import { getRekoStatusSickLeaveTimestamp } from '../../utils/getRekoStatusSickLeaveTimestamp'
+
+export function SelectRekoStatus({
+  statusFromSickLeave,
+  patientId,
+  endDate,
+  rekoStatusTypes,
+}: {
+  statusFromSickLeave: RekoStatus
+  patientId: string
+  endDate: string
+  rekoStatusTypes: RekoStatusType[]
+}) {
+  const [setRekoStatus] = useSetRekoStatusMutation()
+  const emptyRekoStatusId = rekoStatusTypes.find((rekoStatusType) => rekoStatusType.name === 'Ingen')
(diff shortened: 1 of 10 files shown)


apps/rehabstod/src/components/SelectRekoStatus/SelectRekoStatus.tsx
```diff
@@ -0,0 +1,43 @@
...
+    }
+  }
+
+  return (
+    <SelectFilter
+      value={savedRekoStatus}
+      onChange={(id) => handleSetRekoStatus(id)}
+      options={rekoStatusTypes}
+      hideDefaultValue
+      description="Beskrivning"
```
ChrilleSigerhed, line 39: Is this description displayed? If so, maybe we should change it to a more userfriendly message.

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.tsx
```diff
@@ -146,7 +150,7 @@ export function TableBodyRows({
               tabIndex={0}
               onKeyDown={({ code, currentTarget }) => {
                 if (['Enter', 'Space'].includes(code)) {
-                  navigateToPatient(sickLeave.encryptedPatientId)
+                  navigateToPatient(sickLeave)
```
ChrilleSigerhed, line 153: Since we are only using the attributes encryptedPatientId and rekoStatus, maybe we should change the method signature and send these values into the method, instead of the whole sickLeave object?

apps/rehabstod/src/utils/getRekoStatusSickLeaveTimestamp.ts
```diff
@@ -0,0 +1,4 @@
+import { format, isAfter } from 'date-fns'
+
+export const getRekoStatusSickLeaveTimestamp = (endDate: string) =>
```
ChrilleSigerhed, line 3: Nice that you made this into a reusable util and added tests to verify the logic.

ChrilleSigerhed (approved):
Great work, added some small minor comments to consider before merging.

apps/rehabstod/src/components/SelectRekoStatus/SelectRekoStatus.tsx
mhornfeldt: Yes I've asked requirements but haven't gotten an answer yet.

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.tsx
mhornfeldt: I think it's better to leave it generic in case we want to send this as a prop later then the function can be written generic and not so specific.

PR #987, opened 2023-06-26
INTYGFV-15580: Added light prop to ids components

mhornfeldt:


apps/rehabstod/src/pages/Patient/Patient.tsx
```diff
@@ -33,25 +32,14 @@ export function Patient() {
...
-              <div className="w-64">
-                <SelectRekoStatus
-                  endDate={currentSickness.slut}
-                  patientId={firstCertificate.patient.id}
-                  statusFromSickLeave={state.rekoStatus}
-                  rekoStatusTypes={populatedFilters ? populatedFilters.rekoStatusTypes : []}
-                />
-              </div>
-            )}
+            <PatientRekoStatus currentSickLeaves={currentSickLeaves} earlierSickLeaves={earlierSickLeaves} />
```
ChrilleSigerhed, line 42: Cool that u added this component aswell. A minor suggestion that you are free to ignore if u disagree, but since we already have the logic related to the certificates we need, should we use it instead of sending in the whole list? This might affect the reusability of the component, so I let you decide if it makes sense or not.

```javascript
<PatientRekoStatus currentSickLeaves={currentSickness} earlierSickLeaves={firstCertificate} />
```

ChrilleSigerhed (approved):
Great that you added the new light prop to IDS! Left a minor suggestion regarding the PatientRekoStatus component.

apps/rehabstod/src/pages/Patient/Patient.tsx
mhornfeldt: We don't have the logic for the first certificate of the earlierSickLeaves so that's why I placed it in the component to remove some of that logic since it's only used by PatientRekoStatus. The firstCertificate here is the first certificate of the first active sick leave, currentSickness

PR #988, opened 2023-06-26
INTYGFV-15737: Added sticky headers to sick leave view and patient view.

ChrilleSigerhed:


SamuelTilly (commented):
While the solution works, it's not optimal. We could pass parameters from, for example, `FloatingScroll` down to thead by using contexts. But I'm not sure if we can avoid creating divs inside the table. 

When assigning attributes to elements, it's often better to first assign them to a state and then apply the state to our element. This is to avoid loosing those attributes when react decides it wants to re-render. Something like:

```tsx
import { ReactNode, useEffect, useRef, useState } from 'react'
import { classNames } from '../../utils/classNames'

export function FixedTableHeader({ children }: { children: ReactNode }) {
  const ref = useRef<HTMLTableSectionElement>(null)
  const [fixed, setFixed] = useState(false)

  useEffect(() => {
    const handleScroll = () => {
      const parentRect = ref.current?.parentElement?.getBoundingClientRect()
      if (parentRect && ref.current) {
        const { top, bottom } = parentRect
        setFixed(top < 0 && bottom > 50)
      }
    }
    window.addEventListener('scroll', handleScroll, true)
    return () => {
      window.removeEventListener('scroll', handleScroll, true)
    }
  })

  return (
    <>
      {fixed && (
        <thead className="fixed top-0 z-20 overflow-hidden" style={{ width: '1024px' }}>
          {children}
        </thead>
      )}
      <thead ref={ref} className={classNames(fixed && 'invisible')}>
        {children}
      </thead>
    </>
  )
}
```

This doesn't solve the horizontal scroll however.

SamuelTilly (approved):

PR #989, opened 2023-06-27
Feature/intygfv 15688

SamuelTilly:


ChrilleSigerhed (approved):
Awsome work! Great that u managed to solve this using 'react-aria' aswell!

We discussed the possibility to change the opacity on the draged object but we came to the conclusion that it may be browser related and not something we can control.

Excellent work! :-)

PR #990, opened 2023-06-27
INTYGFV-15360: Added LU table for patient view

mhornfeldt:


ChrilleSigerhed (changes requested):
Great work!

Im not done with the review, but this is what I found so far after manually testing the new table.

After clicking on "Visa" in LU-view, the IDSTabPanel gets highlighted. This might be confusing for our users, since they are still on the LU-page. Not sure if this is issue is related to IDSTabs.

image

apps/rehabstod/src/components/Table/utils/filterTableColumns.ts
```diff
@@ -6,11 +6,18 @@ export const filterTableColumns = (
...
-  excludedColumns?: string[]
+  excludedColumns?: string[],
+  isPatientView?: boolean
 ) =>
   columns
     .filter(({ visible }) => visible || !onlyVisible)
     .filter(({ name }) => !(showPersonalInformation === false && (name === 'Namn' || name === 'Personnummer')))
     .filter(({ name }) => !(isDoctor && name === 'Läkare'))
     .filter(({ name }) => !(!srsActivated && name === 'Risk'))
+    .filter(({ name }) => !(isPatientView && (name === 'Namn' || name === 'Personnummer' || name === 'Kön' || name === 'Ålder')))
```
ChrilleSigerhed, line 17: I think this filter has grown alot. Instead of building on top of it, I think it would be a good idea to create a separate filter function, like we already one for filterHiddenColumns for example.

This will increase the readbility in my opinion and makes the code easier to understand.

apps/rehabstod/src/store/api.ts
```diff
@@ -142,6 +142,14 @@ export const api = createApi({
       }),
       providesTags: ['SickLeavePatient'],
     }),
+    getLUCertificatesForPatient: builder.query<LUCertificatesInfo, { encryptedPatientId: string }>({
+      query: (request) => ({
+        url: 'certificate/lu/person',
+        method: 'POST',
+        body: request,
+      }),
+      providesTags: ['LUCertificatesPatient'],
```
ChrilleSigerhed, line 151: We never cache the value from getSickLeavePatient by adding the ```keepUnusedDataFor```-flag. I think we should do the same for this request, so that we always get "latest and greatest"

ChrilleSigerhed (changes requested):
Awsome work, the code is easy to understand and follows a clear pattern, very impressive!. 

I left some comments regarding readability and RTK-query.

apps/rehabstod/src/components/Table/utils/filterTableColumns.ts
mhornfeldt: Good idea! I'll fix this and the comment below in the code I've already written for AG to avoid merge conflicts.

ChrilleSigerhed (approved):
To avoid conflics changes will be handled in AG-PR.

PR #993, opened 2023-06-27
INTYGFV-15379: Added AG certificates in patient view

mhornfeldt:


SamuelTilly (approved):
The endpoint 'certificate/ag/person' could be added to `src\mocks\handlers\handlers.ts`, other than that it looks excellent.

PR #994, opened 2023-06-28
Fix minor issue with navigation from LU to patient view

SamuelTilly:
"Läkarutlåtande" should still be highlighted when navigating from there to patient.

image


ChrilleSigerhed (approved):
Nice catch, well done!

PR #996, opened 2023-06-28
Simplify API by reducing tags

SamuelTilly:
Instead of having unique tags per endpoint, we can have common tags that get invalidated for multiple endpoints. For example, if we have endpoints that need to be invalidated when the user settings get updated, we can simply tag them together with "User".

### DIFF
--- a/apps/rehabstod/src/store/api.ts
+++ b/apps/rehabstod/src/store/api.ts
@@ -97,7 +97,6 @@ export const api = createApi({
         url: 'sickleaves/active',
         method: 'POST',
         body: request,
-        providesTags: ['SickLeaves'],
       }),
       providesTags: ['User'],
     }),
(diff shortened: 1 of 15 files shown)


apps/rehabstod/src/store/api.ts
```diff
@@ -99,15 +99,14 @@ export const api = createApi({
         body: request,
         providesTags: ['SickLeaves'],
```
ChrilleSigerhed, line 100: should this tag be updated aswell?

ChrilleSigerhed (approved):
Nice, left a comment about one tag that might still need to be updated.

apps/rehabstod/src/store/api.ts
SamuelTilly: nice catch! it should :)

PR #1000, opened 2023-06-29
INTYGFV-15757: Added option to ignore input length for FormattedNumbe…

ChrilleSigerhed:
…rInput.


apps/rehabstod/src/components/Form/FormattedNumberInput.tsx
```diff
@@ -26,6 +28,9 @@ export function FormattedNumberInput({
   }
 
   function maxLengthReached(event: React.KeyboardEvent<HTMLInputElement>) {
+    if (ignoreInputLimit) {
+      return false
+    }
     return event.currentTarget.value.length === 2 && event.key !== 'Backspace' && selection?.type !== 'Range'
```
SamuelTilly, line 34: Could we instead get the length from the max attribute?

`event.currentTarget.value.length === `${max}`.length`

SamuelTilly (changes requested):

apps/rehabstod/src/components/Form/FormattedNumberInput.tsx
ChrilleSigerhed: Oh, thats brilliant! Great suggestion, updated the PR!

SamuelTilly (approved):
Nice!

PR #1003, opened 2023-06-30
INTYGFV-15740: Added contact info to questions

mhornfeldt:


packages/common/src/types/question.ts
```diff
@@ -17,6 +17,7 @@ export interface Question {
   lastUpdate: string
   links: ResourceLink[]
   lastDateToReply?: string
+  contactInfo: string[]
```
SamuelTilly, line 20: This is used as `question.contactInfo &&` suggesting that this could be undefined or null. Please update interface to reflect that. `contactInfo?: string[]` or `contactInfo: string[] | null`

SamuelTilly (approved):
One minor issue, otherwise great.

PR #1004, opened 2023-06-30
INTYGFV-15759: Patient LU is using the wrong slice for modifying table columns

SamuelTilly:
Could probably be improved to reduce code duplication, but let's fix the bug first :)


apps/rehabstod/src/pages/CurrentSickLeaves/components/ModifySicknessTableColumns.tsx
```diff
@@ -3,8 +3,8 @@ import { ModifyTableColumns } from '../../../components/Table/ModifyTableColumns
 import { filterTableColumns } from '../../../components/Table/utils/filterTableColumns'
 import { useGetSickLeavesFiltersQuery, useGetUserQuery } from '../../../store/api'
 import { useAppDispatch, useAppSelector, useUpdateUserPreferences } from '../../../store/hooks'
-import { allSickLeaveColumns, sickLeaveColumnsString } from '../../../store/slices/sickLeaveTableColumns.selector'
-import { hideColumn, moveColumn, setColumnDefaults, showColumn } from '../../../store/slices/sickLeaveTableColumns.slice'
+import { allSickLeaveColumns, sickLeaveColumnsString } from '../../../store/slices/sjukfallTableColumns.selector'
```
mhornfeldt, line 6: We shouldn't use Swedish terms in the code

mhornfeldt: We shouldn't adjust to the terms that the previous client has used. If we start using Swedish terms in some cases it's easy that it starts being used more. There's no need to have Swedish terms for these files even if the preference keys are Swedish due to the database being shared.

ChrilleSigerhed: I also feel like using Swedish terms is something we should try to avoid, especially regarding file names. Its my understanding that we already established a translation for these terms, and that this would break the current pattern for file names.

SamuelTilly: Alright, I'll change the filenames back, but could I still keep the generated functions with the desire of generating `tableColumnSlice` easier in the future? https://github.com/sklintyg/frontend/pull/1004/files#diff-fabe1a2336a748ff0695513e8a8c8551c80a2c4a0ae8bb7a982b040f81e8ab0dR106-R114

BTW, is LU really a translation? 😉

mhornfeldt: It's not a translation but it's been decided during a design discussion for the LU table. LU and AG are used in the backend which is why it was chosen.

SamuelTilly: One could argue that "sjukfallTableColumns" is also decided by the backend, but I was just a bit jokingly fastidious in my previous comment about the acronym. I have updated the PR and I believe everything should be translated now 😃

ChrilleSigerhed (approved):
Awsome, I think this looks great!

PR #1020, opened 2023-07-04
INTYGFV-15763: Updated logging for SRS

mhornfeldt:


SamuelTilly (dismissed):

mww-sopra-steria (changes requested):
Looks great! Could you add some tests to the new component and middleware behaviours?

ChrilleSigerhed (commented):
Great job! I Like how you solved the issue with controlling if the SrsPanel is active!

Should we add the new feature launchFromOrigin when the user opens the certificate from rs and open the SrsPanel when that happends?

mhornfeldt:
> Great job! I Like how you solved the issue with controlling if the SrsPanel is active!
> 
> Should we add the new feature launchFromOrigin when the user opens the certificate from rs and open the SrsPanel when that happends?

I've added a new jira for this.

ChrilleSigerhed (approved):

PR #1023, opened 2023-07-05
INTYGFV-15767: Updated icons for global alert

mhornfeldt:


apps/rehabstod/src/components/Layout/Layout.tsx
```diff
@@ -15,7 +15,9 @@ export function Layout() {
     <div className="flex min-h-screen flex-col">
       <LayoutHeader />
       <main className="flex-1">
-        {config && config.banners.length > 0 && <GlobalAlert>{config.banners[0].message}</GlobalAlert>}
+        {config && config.banners.length > 0 && (
+          <GlobalAlert priority={config.banners[0].priority}>{config.banners[0].message}</GlobalAlert>
+        )}
```
SamuelTilly, line 20: Should we only display the first available banner if there are more than one? Looking at the old client's code; it looks like it displays them all.

apps/rehabstod/src/components/GlobalAlert/GlobalAlert.tsx
```diff
@@ -1,10 +1,13 @@
...
-import { IDSAlertGlobal, IDSIconAttention } from '@frontend/ids-react-ts'
+import { IDSAlertGlobal, IDSIconAttention, IDSIconInformation, IDSIconWarning } from '@frontend/ids-react-ts'
+import { BannerPriority } from '../../schemas'
 
-export function GlobalAlert({ children }: { children: ReactNode }) {
+export function GlobalAlert({ children, priority }: { children: ReactNode; priority: BannerPriority }) {
   return (
     <IDSAlertGlobal headline="Driftmeddelande">
-      <IDSIconAttention title="informationsikon" />
+      {priority === BannerPriority.LOW && <IDSIconInformation data-testid="LOW_ICON" />}
```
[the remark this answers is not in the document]
mhornfeldt: This doesn't work for how the IDS-component is rendered. We need to have it inside the return statement and not in a function. From what I remember we can't have multiple ?:-expressions either.

apps/rehabstod/src/components/Layout/Layout.tsx
mhornfeldt: I can add it but test said that only one banner could be added at a time in IA and in the requirements meeting we also talked about that we don't add more than one and don't want to show more than one.

SamuelTilly: Alright, we can keep it like this, was just curious :)

SamuelTilly (approved):

PR #1025, opened 2023-07-05
INTYGFV-15771: Updates to statistical overview

mhornfeldt:


ChrilleSigerhed (changes requested):
Nice, I like the new look. One small fix is required for DiagnosisGroup, it seems its lacking some margin/padding wich results in an overflow of the text.

image

ChrilleSigerhed (approved):
Nice!

PR #1026, opened 2023-07-06
INTYGFV-15760: Added global handlers for fetch, Headers, Request & R…

ChrilleSigerhed:
…esponse to enable running vitest locally with mcw & rtq-query.


SamuelTilly:
Nice! Would it be possible to remove `whatwg-fetch` with this change? I believe it serves a similar purpose.

PR #1027, opened 2023-07-07
INTYGFV-15768: Only load user when session is available 

ChrilleSigerhed:


ChrilleSigerhed:
Tests needs to be updated.

mhornfeldt (approved):

PR #1047, opened 2023-07-20
INTYGFV-15744: Reduce noise from test output. Still not able to remov…

ChrilleSigerhed:
…e stdout related to RelativeContainer in <PieChart> https://github.com/recharts/recharts/issues/2166


mhornfeldt (approved):

SamuelTilly:
This is a great start, but the main issue is in `fakerFromSchemaFactory`, when we create `initialData` it gets copied for every result and therefore are nor unique. I created a fix for the function along with some more changes to further reduce noise, https://github.com/sklintyg/frontend/pull/1064.

PR #1049, opened 2023-07-27
INTYGFV-15775: Fixed tests related to useSession

ChrilleSigerhed:


ChrilleSigerhed:
The issue with these tests are related to msw not beeing able to mock different responses for multiple requests to the same endpoint. The problem we get when we try to mock the session-ping request, is that when we mock the response with authenticated false and hasSession false our user object is never retrieved. 

This results in that our logout-hook is never triggered, or that our siths logout URL is never fetched, since they both require the user object.

My opinion is that we should not let the tests alter the code in this way, were we use the hasSession flag to trigger the user request, even tho the actual effect this change would have is very small. Instead we should just add the skip-flag for these tests.

Therefore, consider this a suggested workaround until msw allows mutiple responses for same request, if we feel that the value of these tests are high enought so that we should keep them.

mhornfeldt (approved):

SamuelTilly:
> The issue with these tests are related to msw not beeing able to mock different responses for multiple requests to the same endpoint. The problem we get when we try to mock the session-ping request, is that when we mock the response with authenticated false and hasSession false our user object is never retrieved.
> 
> This results in that our logout-hook is never triggered, or that our siths logout URL is never fetched, since they both require the user object.
> 
> My opinion is that we should not let the tests alter the code in this way, were we use the hasSession flag to trigger the user request, even tho the actual effect this change would have is very small. Instead we should just add the skip-flag for these tests.
> 
> Therefore, consider this a suggested workaround until msw allows mutiple responses for same request, if we feel that the value of these tests are high enought so that we should keep them.

The problem is rather that when we add `skipToken` to a query, we get an `undefined` object back as a result. So in `useSession` the user object will always be undefined when `session.authenticated` is false. And the same goes for `useLogout` when the logout function is called the user object will always be undefined when `session.authenticated` is false, this would result in `fakeLogout` being called instead of SAML.

I fixed the tests in https://github.com/sklintyg/frontend/pull/1061 by letting useLogout use the [queryState ](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequerystate) instead of [useQuery](https://redux-toolkit.js.org/rtk-query/api/created-api/hooks#usequery). and for `useSession` i wait for both user and session requests to finish and then manually override session query data with the [updateQueryData](https://redux-toolkit.js.org/rtk-query/api/created-api/api-slice-utils#updatequerydata) util.

PR #1061, opened 2023-08-07
INTYGFV-15775: Fix issues with logout and session hooks

SamuelTilly:
explanation:
https://github.com/sklintyg/frontend/pull/1049#issuecomment-1667507560

Excuse the many commits, had to rebase since the branch already existed, and we have changed from master to main.

### DIFF
--- a/apps/rehabstod/src/hooks/useLogout.ts
+++ b/apps/rehabstod/src/hooks/useLogout.ts
@@ -1,8 +1,8 @@
 import { useNavigate } from 'react-router-dom'
-import { useFakeLogoutMutation, useGetUserQuery } from '../store/api'
+import { api, useFakeLogoutMutation } from '../store/api'
 
 export function useLogout() {
-  const { data: user } = useGetUserQuery()
+  const { data: user } = api.endpoints.getUser.useQueryState()
   const [fakeLogout] = useFakeLogoutMutation()
   const navigate = useNavigate()
 
--- a/apps/rehabstod/src/hooks/useSession.test.tsx
+++ b/apps/rehabstod/src/hooks/useSession.test.tsx
@@ -1,7 +1,6 @@
 import { act, screen } from '@testing-library/react'
-import { matchRequestUrl, rest } from 'msw'
+import { rest } from 'msw'
 import { Route, Routes } from 'react-router-dom'
-import { vi } from 'vitest'
 import { server } from '../mocks/server'
 import { api } from '../store/api'
 import { store } from '../store/store'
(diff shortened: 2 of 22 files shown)


apps/rehabstod/src/hooks/useLogout.ts
```diff
@@ -1,8 +1,8 @@
 import { useNavigate } from 'react-router-dom'
-import { useFakeLogoutMutation, useGetUserQuery } from '../store/api'
+import { api, useFakeLogoutMutation } from '../store/api'
 
 export function useLogout() {
-  const { data: user } = useGetUserQuery()
+  const { data: user } = api.endpoints.getUser.useQueryState()
```
SamuelTilly, line 5: This will load existing user data, but will never trigger a fetch. `useGetUserQuery` will return `{data: undefined}` when `session.authenticated` is false.

apps/rehabstod/src/hooks/useSession.test.tsx
```diff
@@ -25,40 +28,53 @@ describe('useSession', () => {
...
+
+    act(() => {
+      store.dispatch(
+        api.util.updateQueryData('getSessionPing', undefined, () => ({
+          hasSession: false,
+          secondsUntilExpire: 0,
+          authenticated: false,
+        }))
+      )
+    })
```
SamuelTilly, line 77: Manually override session data to simulate an end-of-life session.

apps/rehabstod/src/hooks/useSession.ts
```diff
@@ -1,21 +1,28 @@
...
   })
   const { data: user, isLoading: isLoadingUser } = useGetUserQuery()
 
   useEffect(() => {
     if (user && user.pdlConsentGiven === false && isUninitialized) {
       giveConsent({ pdlConsentGiven: true })
     }
-    if (user && session && !session.authenticated) {
-      logout()
+    if (session) {
```
SamuelTilly, line 19: `user` from `useGetUserQuery` will always be undefined when `session.authenticate` is false, so we can not look at the user object to determine if logout should be called.

SamuelTilly:
Tests stopped working after merge, I'll investigate further.

apps/rehabstod/src/hooks/useSession.ts
ChrilleSigerhed: This will open ``` window.open('/saml/logout', '_self')``` when the user opens rehabstod, since they are not authenticated yet.

ChrilleSigerhed (changes requested):
Great work, I think we are getting closer to solving this issue!

I left a comment regarding removing the user-object.

ChrilleSigerhed (approved):
Awsome!

PR #1068, opened 2023-08-08
Use PrintButton component

SamuelTilly:


apps/rehabstod/src/components/PrintButton/PrintButton.tsx
```diff
@@ -2,7 +2,7 @@ import { IDSButton } from '@frontend/ids-react-ts'
 
 export function PrintButton() {
   return (
-    <IDSButton onClick={() => window.print()} className="mb-3 whitespace-nowrap">
+    <IDSButton size="l" onClick={() => window.print()} className="mb-3 whitespace-nowrap">
```
SamuelTilly, line 5: Does not look correct right now but should be fixed in future IDS release

ChrilleSigerhed (approved):

PR #1081, opened 2023-08-17
INTYGFV-15805: Fixed problems with reko status resetting when switchi…

mhornfeldt:
…ng tabs in patient view


SamuelTilly (commented):
While this work for the scenario described in the ticket. We won't get rekostatus when we navigate from "LUCertificate" to Patient view and then click on sickleave. Could we add the rekostatus to '/api/sjukfall/patient' endpoint and use that state instead?

SamuelTilly (approved):

PR #1082, opened 2023-08-17
INTYGFV-15804: Fixed cache problem related to LU certificates

mhornfeldt:


apps/rehabstod/src/store/api.ts
```diff
@@ -60,7 +60,7 @@ export const api = createApi({
         try {
           await queryFulfilled
         } catch {
-          dispatch(api.util.invalidateTags(['User']))
+          dispatch(api.util.invalidateTags(['User', 'LUCertificates']))
```
SamuelTilly, line 63: This should only invalidate 'User'

apps/rehabstod/src/store/api.ts
```diff
@@ -46,7 +46,7 @@ export const api = createApi({
         method: 'POST',
         body: { id: vardenhet.id },
       }),
-      invalidatesTags: ['User'],
+      invalidatesTags: ['User', 'LUCertificates'],
```
SamuelTilly, line 49: This should only invalidate 'User'

apps/rehabstod/src/store/api.ts
```diff
@@ -148,6 +148,7 @@ export const api = createApi({
         method: 'POST',
         body: request,
       }),
+      providesTags: ['LUCertificates'],
```
SamuelTilly, line 151: `providesTags: ['User']`

apps/rehabstod/src/store/api.ts
```diff
@@ -161,6 +162,7 @@ export const api = createApi({
     >({
       keepUnusedDataFor: 0,
       query: () => 'lu/filters',
+      providesTags: ['LUCertificates'],
```
SamuelTilly, line 165: `providesTags: ['User']`

SamuelTilly (changes requested):
Similar to `getSickLeave`, the LU endpoints should provide "User" tag, this way they get invalidated together with every other endpoint that is depending on the current user. Also, `getPatientLUCertificates` should probably be tagged with 'Patient', same as `getPatientSickLeaves`.

PR #1085, opened 2023-08-21
Rehabstod - API code splitting

SamuelTilly:
Split `api.ts` into multiple files for easier overview and potential smaller bundle size. https://redux-toolkit.js.org/rtk-query/usage/code-splitting


ChrilleSigerhed (approved):
Great idea so separate the different calls like this, lets make an effort to keep the same structure for our upcoming applications

PR #1093, opened 2023-08-23
INTYGFV-15809: Spinner is not hidden after certificate gets deleted

SamuelTilly:


apps/webcert/src/store/certificate/certificateMiddleware.ts
```diff
@@ -241,12 +241,12 @@ const handleDeleteCertificateSuccess: Middleware<Dispatch> =
   ({ dispatch }) =>
   () =>
   (action: AnyAction): void => {
+    dispatch(hideSpinner())
```
ChrilleSigerhed, line 244: Does it make more sense to move this dispatch to line 252, and hide the spinner after the other dispatches has finished?

ChrilleSigerhed (approved):
See comment!

Also, would be great if we could add a test to verify that the spinner is indeed hidden correctly!

apps/webcert/src/store/certificate/certificateMiddleware.ts
SamuelTilly: I don't think so, at least not in any meaningful way, it would be milliseconds for every dispatch to fire and update state

SamuelTilly:
> See comment!
> 
> Also, would be great if we could add a test to verify that the spinner is indeed hidden correctly!

Good point, added a test and moved the dispatch 👍

PR #1094, opened 2023-08-25
INTYGFV-15808: No empty tooltip for diagnosis

SamuelTilly:


ChrilleSigerhed (approved):
Nice! Would be nice if we could also add a test to verify this new behaviour!

PR #1099, opened 2023-08-28
INTYGFV-15820: Updated how rekostatus is retrieved for patient view

mhornfeldt:


apps/rehabstod/src/store/sickLeaveApi.ts
```diff
@@ -109,6 +110,14 @@ const sickLeaveApi = api.injectEndpoints({
         }
       },
     }),
+    getRekoStatusForPatient: builder.query<RekoStatus, { patientId: string; endDate: string; startDate: string }>({
+      keepUnusedDataFor: 0,
```
SamuelTilly, line 114: Perhaps we could use the cache instead? We could add a tag with specific identifier that gets invalidated with setRekoStatus.

setRekoStatus:
```typescript
invalidatesTags: (result, error, {patientId}) => [{type: 'RekoStatus', id: patientId}],
```

getRekoStatusForPatient:
```typescript
providesTags: (result, error, {patientId}) => [{ type: 'RekoStatus', id: patientId }],
```

apps/rehabstod/src/pages/Patient/components/SelectRekoStatus/SelectRekoStatus.tsx
```diff
@@ -1,42 +1,44 @@
...
       setRekoStatus({ patientId, status: type, sickLeaveTimestamp, filter })
       updateSavedRekoStatus(type.id)
-      dispatch(updateRekoStatusId(type.id))
     }
   }
 
+  useEffect(() => {
+    updateSavedRekoStatus(rekoStatusFromDatabase?.status.id)
+  }, [rekoStatusFromDatabase])
+
```
SamuelTilly, line 41: Not sure if we need to store `rekostatus` in a local state. We could use the state that the API provides.

```typescript
const savedRekoStatus = rekoStatusFromDatabase?.status.id || emptyRekoStatus?.id
```

apps/rehabstod/src/pages/Patient/components/SelectRekoStatus/SelectRekoStatus.tsx
```diff
@@ -1,42 +1,44 @@
...
   patientId: string
   endDate: string
+  startDate: string
   rekoStatusTypes: RekoStatusType[]
   description: string
   disabled?: boolean
 }) {
   const [setRekoStatus] = useSetRekoStatusMutation()
-  const { rekoStatusId } = useAppSelector((rootState) => rootState.sickLeave)
+  const { data: rekoStatusFromDatabase } = useGetRekoStatusForPatientQuery({ patientId, endDate, startDate })
```
SamuelTilly, line 24: "FromDatabase" is a bit verbose, I guess everything from the API is stored in a database.

apps/rehabstod/src/pages/Patient/components/PatientRekoStatus.tsx
```diff
@@ -16,8 +12,6 @@ export function PatientRekoStatus({
   isDoctor: boolean
 }) {
   const { data: populatedFilters } = useGetSickLeavesFiltersQuery()
-  const { state } = useLocation()
```
SamuelTilly, line 19: This state can be removed from `apps\rehabstod\src\pages\CurrentSickLeaves\components\TableBodyRows.tsx`as well.

SamuelTilly (changes requested):
Awesome that this could be fixed! found some minor things that could be changed

apps/rehabstod/src/store/sickLeaveApi.ts
mhornfeldt: I remember Matias mentioned at a meeting that we shouldn't use the cache at all for the patientview.

SamuelTilly: Alright, let's keep the `keepUnusedDataFor`. I think we still need to invalidate `getRekoStatusForPatient` when `setRekoStatus` gets called.

mhornfeldt: This gives a ts error

SamuelTilly: Probably missing 'RekoStatus' in the list of tags in `api.ts` ?

SamuelTilly (approved):

PR #1103, opened 2023-08-30
INTYGFV-15802: Added number column to sick leave and lu table

mhornfeldt:


apps/rehabstod/src/components/Table/tableHeader/TableHeaderCell.tsx
```diff
@@ -4,15 +4,25 @@ import { TooltipTrigger } from '../../Tooltip/TooltipTrigger'
...
+export function TableHeaderCell({
+  column,
+  width,
+  sticky,
+  unSortable,
+}: {
+  column: string
+  width?: string
+  sticky?: 'left' | 'top' | 'right'
+  unSortable?: boolean
```
SamuelTilly, line 16: I think changing this to "sortable" would be easier to reason about. It could also have a default value:

```typescript
{sortable = true}: {sortable?: boolean}
```
```typescript
tabIndex={sortable ? 0 : undefined}
```

apps/rehabstod/src/components/Table/tableHeader/TableHeaderCell.tsx
```diff
@@ -27,9 +37,9 @@ export function TableHeaderCell({ column, width, sticky }: { column: string; wid
               ;(currentTarget.nextElementSibling as HTMLElement).focus()
             }
           }}
-          onClick={() => sortOnColumn(column)}
+          onClick={unSortable ? () => {} : () => sortOnColumn(column)}
```
SamuelTilly, line 40: `sortOnColumn` gets called in `onKeyDown` as well.

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.test.tsx
```diff
@@ -136,7 +136,7 @@ it('Should render all sickleave columns', async () => {
   )
 
   expect(await screen.findAllByRole('row')).toHaveLength(10)
-  expect(screen.getAllByRole('row')[0].children).toHaveLength(14)
+  expect(screen.getAllByRole('row')[0].children).toHaveLength(15)
```
SamuelTilly, line 139: We could get the number of columns from `Object.keys(SickLeaveColumn).length` the same way we do in `PatientTableBody`. But this is not super important, just something I noticed.

SamuelTilly (changes requested):

apps/rehabstod/src/components/Table/tableHeader/TableHeaderCell.tsx
mhornfeldt: Since you can't tab to it I don't think this is possible to call but I can remove it

apps/rehabstod/src/pages/CurrentSickLeaves/components/TableBodyRows.test.tsx
mhornfeldt: 15 is not the length so I will leave it as it is. I think it will be more confusing if we start adding subtractions in several places

apps/rehabstod/src/components/Table/tableHeader/TableHeaderCell.tsx
SamuelTilly: Good point

SamuelTilly (approved):

PR #1108, opened 2023-09-04
INTYGFV-15828: Add certificate list filters

SamuelTilly:


apps/minaintyg/src/components/FilterAccordion/FilterAccordion.tsx
```diff
@@ -0,0 +1,17 @@
+import { IDSIconChevron } from '@frontend/ids-react-ts'
+import { ReactNode } from 'react'
+
+export function FilterAccordion({ title, children }: { title: string; children: ReactNode }) {
```
mhornfeldt, line 4: Can be made to a general component. Only the text "Filtrera lista" is unique for Filter.

apps/minaintyg/src/components/FilterAccordion/FilterAccordion.tsx
```diff
@@ -0,0 +1,17 @@
...
+  return (
```
mhornfeldt, line 5: Test is missing

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateListFilter.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+  server.use(
+    rest.get('/api/filter-certificate', (_, res, ctx) =>
+      res(ctx.status(200), ctx.json(fakerFromSchema(certificateFilterOptionsSchema)(options)))
+    )
+  )
+})
+
+describe('Options from API', () => {
+  beforeEach(() => {
+    render(
```
mhornfeldt, line 30: Render can be extracted to function which accepts onSubmit as parameter

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateListFilter.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+      <CertificateListFilter onSubmit={onSubmit} />
+    </Provider>
+  )
+
+  await userEvent.click(screen.getByLabelText('Filtrera'))
+
+  expect(onSubmit).toHaveBeenCalledTimes(1)
+})
+
+it('Should reset filters when "Återställ filter" is pressed', async () => {
```
mhornfeldt, line 60: This test is long and a bit hard to understand. Maybe it can be made easier to understand if some parts were extracted to functions and the test was split up into several tests.

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateListFilter.tsx
```diff
@@ -0,0 +1,33 @@
+import { IDSButton } from '@frontend/ids-react-ts'
```
mhornfeldt, line 1: Tests are missing

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateStatusFilter.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { useAppDispatch, useAppSelector } from '../../../../store/hooks'
+import { update } from '../../../../store/slice/certificateFilter.slice'
+
+export function CertificateStatusFilter({ options }: { options: CertificateStatus[] }) {
+  const id = useId()
+  const { statuses } = useAppSelector((state) => state.certificateFilter)
+  const dispatch = useAppDispatch()
+
+  return (
+    <IDSSelect>
```
mhornfeldt, line 13: Can the select filters be made into a common component? There are a lot of duplicated code looking at the components and the tests

mhornfeldt (commented):
Nice. Some comments

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateListFilter.test.tsx
SamuelTilly: Rewrote it using `.each` to separate the filters, hopefully more readable now.

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateStatusFilter.tsx
SamuelTilly: I moved the general `Select` component from rehabstod into `@frontend/components` together with its Tooltip dependency, this way both rehabstod and minaintyg can share that component and removes a lot of the duplication. Each filter is similar but updates different parts of the store, has different labels etc. so they all need to be tested individually

apps/minaintyg/src/components/FilterAccordion/FilterAccordion.test.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { FilterAccordion } from './FilterAccordion'
+
+it('Should render as expected', () => {
+  const { container } = render(
+    <FilterAccordion listed={43} total={100} noun="intyg">
+      <span>Test</span>
+    </FilterAccordion>
+  )
+  expect(screen.getByText(`43 av 100 intyg`)).toBeInTheDocument()
+  expect(container).toMatchSnapshot()
```
mhornfeldt, line 11: Tests can be split

apps/minaintyg/src/pages/Certificates/Certificates.tsx
```diff
@@ -18,6 +23,7 @@ export function Certificates() {
         intyg digitalt till Försäkringskassan och Transportstyrelsen. Läkarintyg om arbetsförmåga kan inte skickas digitalt till din
```
mhornfeldt, line 23: Does this component have tests?

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateListFilter.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    rest.get('/api/certificate/filters', (_, res, ctx) =>
+      res(ctx.status(200), ctx.json(fakerFromSchema(certificateFilterOptionsSchema)(options)))
+    )
+  )
+})
+
+describe('Options from API', () => {
+  it.each(['Status', 'Mottagning', 'Intygstyp', 'År'])('Should have corret %s option', async (name) => {
+    const { container } = render(
+      <Provider store={store}>
```
mhornfeldt, line 31: Can be extracted into common function to reduce code in each test and make them easier to read

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateStatusFilter.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { render, screen } from '@testing-library/react'
```
mhornfeldt, line 1: Nice tests that are easy to read!

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateStatusFilter.tsx
```diff
@@ -0,0 +1,18 @@
+import { Select } from '@frontend/components'
+import { CertificateStatus } from '../../../../schema/certificateList.schema'
+import { useAppDispatch, useAppSelector } from '../../../../store/hooks'
+import { update } from '../../../../store/slice/certificateFilter.slice'
+
+export function CertificateStatusFilter({ options }: { options: CertificateStatus[] }) {
+  const { statuses } = useAppSelector((state) => state.certificateFilter)
```
mhornfeldt, line 7: Is it better to send the values for each filter as prop from parent and have the appSelector only in the parent component?

apps/minaintyg/src/schema/certificateListFilter.schema.ts
```diff
@@ -0,0 +1,15 @@
+import { z } from 'zod'
+import { CertificateStatusEnum } from './certificateList.schema'
+
+export const certificateFilterOptionsSchema = z.object({
+  total: z.number().optional(),
```
mhornfeldt, line 5: I've merged this fix to the backend now

apps/rehabstod/src/pages/Patient/components/SelectRekoStatus/SelectRekoStatus.tsx
```diff
@@ -33,12 +33,11 @@ export function SelectRekoStatus({
   }
 
   return (
-    <SelectFilter
+    <Select
       disabled={disabled}
       value={rekoStatus?.status?.id || emptyRekoStatus?.id}
-      onChange={(id) => handleSetRekoStatus(id)}
-      options={rekoStatusTypes}
-      hideDefaultValue
```
mhornfeldt, line 41: Is this functionality still working with the new changes or is it removed from requirements?

mhornfeldt (changes requested):

apps/rehabstod/src/pages/Patient/components/SelectRekoStatus/SelectRekoStatus.tsx
SamuelTilly: It's still working, this was the only component that used `SelectFilter` that needed "hideDefaultValue", when changing it to this new `Select` from `@frontend/components` it's no longer needed and the prop could be deleted from `SelectFilter`

apps/minaintyg/src/pages/Certificates/components/CertificateListFilter/CertificateStatusFilter.tsx
SamuelTilly: The effect would be the same, only we would need to add a property for every filter to pass the value down to the parent and move the tests to the parent for every filter, this would add a bit more code.

One thing I noticed now however is that we should pass the value and not the object from `useSelector` to avoid unnecessary re-renders since `useSelector` evaluates changes with a strict `===` operator.

I'll change them to be:
```typescript
const statuses = useAppSelector((state) => state.certificateFilter.statuses)
```

apps/minaintyg/src/pages/Certificates/Certificates.tsx
SamuelTilly: Yes, tests are available here `apps\minaintyg\src\pages\Certificates\Certificates.test.tsx`

mhornfeldt (approved):

PR #1111, opened 2023-09-11
Feature/vitest workspace

SamuelTilly:


ChrilleSigerhed (approved):
Awsome work! Can we add some samples in our README for running tests in separate files now that we no longer require to use --filter.

PR #1115, opened 2023-09-15
INTYGFV-15835: SRS panel should be active if questions are handled

mhornfeldt:


SamuelTilly (commented):
Should CertificateSidePanel.test.tsx be updated for this new logic?

SamuelTilly (approved):

PR #1129, opened 2023-09-20
INTYGFV-15853: Render certificate content

SamuelTilly:
mi2 mi localtest me_intyg_8ec57d34-aecf-46e9-b132-901664e8f85d


mhornfeldt:
@SamuelTilly From backend can the HTML be delivered without indentation?

SamuelTilly:
> @SamuelTilly From backend can the HTML be delivered without indentation?

Sure, that would be no problem :)

apps/minaintyg/src/mocks/certificateContentMock.ts
```diff
@@ -0,0 +1,148 @@
+export const certificateContentMock = [
```
mhornfeldt, line 1: We should have a talk within the team if the model is changing. The backend solution is built according to the original model we designed.

apps/minaintyg/src/pages/Certificate/CertificatePage.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { faker } from '@frontend/fake'
```
mhornfeldt, line 1: I think we should have a talk within the team about how to write tests in frontend. Right now it feels like snapshot tests are the main technique for testing. My opinion is that snapshot tests can be used when necessary but that they shouldn't replace unit tests.

apps/minaintyg/src/schema/certificate.schema.ts
```diff
@@ -12,9 +12,8 @@ export const certificateSummarySchema = z.object({ label: z.string(), value: z.s
 
 export const certificateTypeSchema = z.object({
   id: z.string(),
-  name: z.string(),
```
ChrilleSigerhed: Since we use the same metadata for both list certificate and read certificate in frontend we need to keep name. It will be used to display the name of the certificate in the list certificate view.

An alternative solution could be that we separate the two different metadata.

ChrilleSigerhed (changes requested):
Nice, almost there now! :)

We also need to change the certificateScheme to the following:

```javascript
export const certificateSchema = z.object({
  certificate: z.object({ content: z.array(certificateContentSchema), metadata: certificateMetadataSchema }),
})
```

apps/minaintyg/src/schema/certificate.schema.ts
SamuelTilly: I moved the name from metadata to type for certificate response using `transformResponse` so that it follows the same schema.

apps/minaintyg/src/pages/Certificate/CertificatePage.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  if (!certificate) {
+    return null
+  }
+
+  return (
+    <>
+      <PageHeading heading={certificate.metadata.type.name}>
+        <DisplayHTML html={certificate.metadata.type.description} />
```
ChrilleSigerhed, line 28: Since description will come in a later stage, maybe we can just return a "TODO" value, or just ignore this for now.

ChrilleSigerhed (approved):
Awsome job, well done! Please remove/change description before merging since it will be implemented in a later stage.

PR #1130, opened 2023-09-21
DEVENV: Added handler for CSRF in request.

ChrilleSigerhed:


SamuelTilly (approved):
Looking good, we can look at moving cookie util into a common shared util package later, since it's shared with rehabstod.

PR #1131, opened 2023-09-21
INTYGFV-15824: login with saml

SamuelTilly:


apps/minaintyg/src/components/Layout/LayoutHeader.tsx
```diff
@@ -21,33 +22,40 @@ export function LayoutHeader() {
 
   return (
```
mhornfeldt, line 23: Component is a bit big. Can it be extracted into several components?

apps/minaintyg/src/hooks/useLogout.ts
```diff
@@ -1,12 +1,25 @@
...
 import { useFakeLogoutMutation } from '../store/testabilityApi'
 
 export function useLogout() {
   const [fakeLogout] = useFakeLogoutMutation()
+  const { data: user } = useGetUserQuery()
   const navigate = useNavigate()
 
   return () => {
-    fakeLogout()
-    navigate('/welcome')
```
mhornfeldt, line 10: Should hooks have tests?

SamuelTilly: Sure, I would prefer to test the effect of a hook, but logout is a bit special. 

Added a test for fake logout. SAML logout is a bit trickier to test since the testing environment has no knowledge about browser path's, so there is no /logout path to navigate to. This would be a good candidate cypress tests.

mhornfeldt (approved):
Nice!

PR #1136, opened 2023-09-27
INTYGFV-15536: Added tooltip for unknown diagnosis

mhornfeldt:


apps/rehabstod/src/components/DiagnosisInfo/DiagnosisInfoCell.tsx
```diff
@@ -11,7 +11,7 @@ export function DiagnosisInfoCell({ diagnosis, biDiagnoses }: { diagnosis?: Diag
           <ul>
             {filteredDiagnosis.map(({ kod, beskrivning }) => (
               <li key={kod}>
-                {kod} - {beskrivning}
+                {kod} - {beskrivning || `Diagnoskod ${kod} är okänd och har ingen beskrivning`}
```
SamuelTilly, line 14: `diagnosisSchema` needs to be updated to `beskrivning: z.string().optional()` to reflect that this property can possibly be undefined.
A test should be added to DiagnosisInfoCell.test.tsx for this condition.

SamuelTilly (changes requested):

SamuelTilly (approved):
Nice!

PR #1139, opened 2023-09-29
INTYGFV-15894: Updated SRS logging

mhornfeldt:


apps/webcert/src/components/srs/panel/SrsPanel.tsx
```diff
@@ -55,13 +62,39 @@ const SrsPanel: React.FC<Props> = ({ minimizedView, isPanelActive }) => {
...
+    if (!hasLoggedMeasuresDisplay) {
+      logMeasuresDisplayed()
+    }
+    const currentRef = ref.current
+    currentRef?.addEventListener('scroll', handleScroll)
+    return () => {
+      currentRef?.removeEventListener('scroll', handleScroll)
     }
-  }, [isPanelActive, dispatch, certificateId, loggedCertificateId])
+  }, [handleScroll, ref, dispatch, logMeasuresDisplayed, hasLoggedMeasuresDisplay])
```
SamuelTilly, line 97: Since the `handleScroll` and `logMeasuresDisplayed` functions are not needed by anything but the `useEffect` we can move the functions in to the `useEffect` and share its dependencies. But this useEffect is getting rather large so perhaps it would be a good idea to move it into it's own file:

```typescript
export function useLogMeasuresDisplayed({ref, measuresRef, isPanelActive}: {ref: RefObject<HTMLElement>, measuresRef: RefObject<HTMLElement>, isPanelActive: boolean}) {
  const dispatch = useAppDispatch()
  const hasLoggedMeasuresDisplay = useSelector(hasLoggedMeasuresDisplayed)

  useEffect(() => {
    const currentRef = ref.current
    const handleScroll = () => {
      if (isPanelActive) {
        dispatch(logSrsInteraction(SrsEvent.SRS_PANEL_ACTIVATED))
        if (!hasLoggedMeasuresDisplay) {
          logMeasuresDisplayed()
        }
      }
    }

    const logMeasuresDisplayed = () => {
      if (measuresRef && measuresRef.current) {
        const isMeasuresVisible = isScrolledIntoView(measuresRef.current, true)
        if (isMeasuresVisible) {
          dispatch(logSrsInteraction(SrsEvent.SRS_MEASURES_DISPLAYED))
          dispatch(updateHasLoggedMeasuresDisplayed(true))
        }
      }
    }

    ReactTooltip.rebuild()
    
    if (!hasLoggedMeasuresDisplay) {
      logMeasuresDisplayed()
    }
    
    currentRef?.addEventListener('scroll', handleScroll)
    return () => {
      currentRef?.removeEventListener('scroll', handleScroll)
    }
  }, [ref, dispatch, hasLoggedMeasuresDisplay, isPanelActive, measuresRef])
}
```

SamuelTilly (approved):
Looking good, only a minor suggestion :)

apps/webcert/src/components/srs/panel/SrsPanel.tsx
mhornfeldt: Good suggestion. But maybe I should leave the tooltip out in its own useEffect since it doesn't belong to logInteraction?

mhornfeldt: If I switch to this I get that ref stays undefined. Maybe I'm calling it wrong in the SrsPanel.

 In SRSPanel I've added: 
useLogSRSInteraction(ref, measuresRef, isPanelActive)

PR #1148, opened 2023-10-06
INTYGFV-15828: Filter certificates

SamuelTilly:
Fixes

* Wrong label on status filter
* Missing unit name on certificate card
* Wrong alert information on empty filtered-list
* Status badges should be sorted so that "new" is always first


mww-sopra-steria (approved):
Nice!

PR #1149, opened 2023-10-08
INTYGFV-15924: Added missing numbering of items in printout of sjukfa…

mikeyOne:
…llslist


apps/rehabstod/src/pages/CurrentSickLeaves/components/PrintTable.tsx
```diff
@@ -27,20 +27,22 @@ export function PrintTable({ sickLeaves, showPersonalInformation }: { sickLeaves
     return null
   }
 
+  const sortedList = sortTableList(sickLeaves, getSickLeavesColumnData)
+
   return (
     <div>
       {sortTableList(sickLeaves, getSickLeavesColumnData)?.map((sickLeave) => (
```
SamuelTilly, line 34: This should be replaced this with the new `sortedList` so that we don't have to sort the list twice :)

SamuelTilly (changes requested):

SamuelTilly (approved):
Nice!

PR #1158, opened 2023-10-18
INTYGFV-15873: Add filter validation

SamuelTilly:
image


ChrilleSigerhed (approved):
Nice, I like how you separated the filter components into individual files. Makes it look clean and easy to follow the code.

As discussed, the remaining tests are handled in a later stage. Well done!

PR #1164, opened 2023-10-23
INTYGFV-15957: Compact print output

SamuelTilly:
image


apps/rehabstod/src/components/Table/TableInfo/TableRowsInfoItem.tsx
```diff
@@ -0,0 +1,12 @@
+import { TableInfoItem } from './TableInfoItem'
```
mhornfeldt, line 1: Tests can be added for this component

mhornfeldt (approved):
Nice layout

PR #1166, opened 2023-10-24
INTYGFV-15886: Handle replaced certificates

SamuelTilly:


apps/minaintyg/src/pages/Certificate/components/CertificateEventsInfo/CertificateEventsInfo.tsx
```diff
@@ -7,10 +9,16 @@ export function CertificateEventsInfo({ events }: { events: CertificateEvent[] }
     <div>
       <div className="font-bold">Senaste händelser</div>
       {events.length === 0 && <span>Inga händelser</span>}
-      {events.map(({ timestamp, description }) => (
+      {events.map(({ timestamp, description, certificateId }) => (
```
ChrilleSigerhed, line 12: We also need to sort the events so that the latest event is displayed at the top. Is that handled? If not, do you think it should be done in backend or frontend?

ChrilleSigerhed: Agreed, I will create a task for it!

ChrilleSigerhed (approved):
Nice, will create a task for the sorting of events.

PR #1168, opened 2023-10-24
INTYGFV-15957: Fixes for compact print output

SamuelTilly:


apps/rehabstod/src/pages/CurrentSickLeaves/utils/sickLeaveColumnFilter.ts
```diff
@@ -0,0 +1,23 @@
+import { TableColumn } from '../../../schemas/tableSchema'
```
mhornfeldt, line 1: Tests are missing

mhornfeldt (changes requested):

apps/rehabstod/src/pages/CurrentSickLeaves/utils/sickLeaveColumnFilter.ts
SamuelTilly: Added a couple of tests

mhornfeldt (approved):

PR #1170, opened 2023-10-25
INTYGFV-16000: Login issues

SamuelTilly:
* Should not render children while loading user in `ProtectedRoute`
* Redirect to `/welcome` when in development mode and not logged in
* Added button from `/welcome` to activate SAML login
* Navigate to `/intyg` when using fake-login


mww-sopra-steria (approved):
Really nice!

PR #1179, opened 2023-10-30
Fix linting errors for webcert tests

SamuelTilly:


ChrilleSigerhed (approved):
Nice, good initiative to update the linting for webcert.

PR #1197, opened 2023-11-08
INTYGFV-16073: Alert messages on failed requests

SamuelTilly:


apps/minaintyg/src/components/error/TechnicalIssueAlert.tsx
```diff
@@ -0,0 +1,11 @@
+import { IDSAlert } from '@frontend/ids-react-ts'
+import { ReactNode } from 'react'
+
+export function TechnicalIssueAlert({ children }: { children: ReactNode }) {
```
mhornfeldt, line 4: Test for component is missing

mhornfeldt (approved):
Approved! If it's not already done, a task can be added to add the errorId to error messages when the id is available from backend

apps/minaintyg/src/components/error/TechnicalIssueAlert.tsx
SamuelTilly: It's tested in both `CertificateListPage` and `CertificatePage` where it's used. :)

PR #1206, opened 2023-11-14
INTYGFV-16089: Added send available function

mhornfeldt:


apps/minaintyg/src/pages/Certificate/components/CertificateActions.tsx
```diff
@@ -1,16 +1,25 @@
 import { IDSButton } from '@frontend/ids-react-ts'
 import { useState } from 'react'
 import { useNavigate } from 'react-router-dom'
-import { CertificateRecipient } from '../../../schema/certificate.schema'
+import { AvailableFunction, AvailableFunctionsTypeEnum, CertificateRecipient } from '../../../schema/certificate.schema'
 import { CertificateSentDialog } from './CertificateSentDialog/CertificateSentDialog'
 
-export function CertificateActions({ recipient }: { recipient?: CertificateRecipient | null }) {
+export function CertificateActions({
```
ChrilleSigerhed, line 7: Since we are introducing print aswell, should we change the name of this component to CertificateSendActions?

ChrilleSigerhed (approved):
Nice, well done!

apps/minaintyg/src/pages/Certificate/components/CertificateActions.tsx
mhornfeldt: I assumed the thought was that print will be placed in this component also since it's placed directly in CertificatePage

ChrilleSigerhed: Alright! Makes sense

PR #1214, opened 2023-11-17
INTYGFV-16082: Log errors

SamuelTilly:


ChrilleSigerhed:
If I produce a CLIENT_ERROR, the errorId presented to the user matches with the one we send to log/error. But if I block the request/take down a service, it does not:

image

apps/minaintyg/src/schema/error.schema.ts
```diff
@@ -2,3 +2,24 @@ import { z } from 'zod'
 
 export const ErrorType = z.enum(['unknown', 'login-failed', 'unavailable', 'logged-out', 'not-found'])
 export type ErrorTypeEnum = z.infer<typeof ErrorType>
+
+export const ErrorCode = z.enum([
+  'CLIENT_ERROR',
+  'LOGIN_FAILED',
+  'UNKNOWN_INTERNAL_ERROR',
+  'FETCH_ERROR',
```
ChrilleSigerhed, line 10: Could ```FETCH_ERROR``` instead be called ```SERVICE_UNAVAILABLE_ERROR```?

ChrilleSigerhed (approved):
Awsome, really well done! Really good tests aswell!

apps/minaintyg/src/schema/error.schema.ts
SamuelTilly: FETCH_ERROR is from redux, could remap them but would rather not if not necessary 
image

PR #1229, opened 2023-11-27
INTYGFV-16119: Added endpoint api/info to get global banners from intygsadmin.

ChrilleSigerhed:
image


mikeyOne (approved):
Looks good

apps/minaintyg/src/components/Layout/GlobalBanner/GlobalBanner.tsx
```diff
@@ -0,0 +1,13 @@
+import { IDSAlertGlobal, IDSIconAttention, IDSIconInformation, IDSIconWarning } from '@frontend/ids-react-ts'
+import { Banner, BannerPriority } from '../../../schema/info.schema'
+
+export function GlobalBanner({ banner }: { banner: Banner }) {
```
mhornfeldt, line 4: Shouldn't this be a common component between apps and not placed in MI? Maybe there's code from RS that can be reused?

apps/minaintyg/src/schema/info.schema.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const bannerSchema = z.object({
+  content: z.string(),
+  type: bannerTypeEnum,
+})
+export const infoSchema = z.object({
+  banners: z.array(bannerSchema),
+})
+
+export type Info = z.infer<typeof infoSchema>
```
mhornfeldt, line 19: Should info be named something less general like InformationResponse or similar?

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
mhornfeldt, line 1: Should this be included in the pr?

apps/minaintyg/src/components/Layout/GlobalBanner/GlobalBanner.tsx
ChrilleSigerhed: There is a difference in BannerPriority that led me to implement a new Component.

apps/minaintyg/src/schema/info.schema.ts
ChrilleSigerhed: Good idea!

PR #1231, opened 2023-11-27
INTYGFV-16049-fix: Scroll position should not be saved on CertificatePage

mikaelaromild1:


apps/minaintyg/src/pages/Certificate/CertificatePage.tsx
```diff
@@ -25,6 +25,7 @@ export function CertificatePage() {
 
   return (
     <>
+      <ScrollRestoration />
```
mikaelaromild1, line 28: Not sure if this is the best way to solve it or if it should be moved, but it seems to work.

mhornfeldt (approved):

apps/minaintyg/src/pages/Certificate/CertificatePage.tsx
mhornfeldt: If it works it seems like a good fix! Did you check so you don't get any errors in the console?

PR #1245, opened 2023-12-11
INTYGFV-16161: Handle empty table header columns

SamuelTilly:


mhornfeldt (approved):
Nice. This can be done in backend also for ViewTable but it requires more changes than in the frontend. But didn't this task include updating some elements in the table body to th also?

SamuelTilly:
> Nice. This can be done in backend also for ViewTable but it requires more changes than in the frontend. But didn't this task include updating some elements in the table body to th also?

Yeah, that is correct, this would only solve the issue with empty th.

PR #1257, opened 2024-01-03
INTYGFV-16246: Removed function .at which is not supported by older b…

mhornfeldt:
…rowsers


mikaelaromild1 (approved):
Well done, good catch! :)

PR #1261, opened 2024-01-09
Update redux

SamuelTilly:


apps/rehabstod/src/hooks/useSession.test.tsx
```diff
@@ -1,81 +1,63 @@
...
-describe('useSession', () => {
-  beforeEach(() => {
-    vi.useFakeTimers()
-  })
-
-  afterEach(() => {
-    vi.restoreAllMocks()
-  })
-
-  it('Should check session every 30 seconds', async () => {
```
mhornfeldt, line 31: Should this test really be removed?

mhornfeldt (approved):

apps/rehabstod/src/hooks/useSession.test.tsx
SamuelTilly: I would prefer to have it working, but was not able to. Not sure what causes it, but I think one test can be sacrificed for upgrading Redux.

PR #1275, opened 2024-01-31
INTYGFV-16320: Certificate type id should be upper case

mhornfeldt:


ChrilleSigerhed (approved):
This could be handled from backend aswell, but I think this is fine!

mhornfeldt:
> This could be handled from backend aswell, but I think this is fine!

We discussed both solutions before implementing. Our strategy was to not manipulate data in the backend to satisfy frontend requirements.

PR #1280, opened 2024-02-19
INTYGFV-16532: Refactor pagination component

SamuelTilly:
Fix sonarqube issues and simplify pagination component


ChrilleSigerhed (approved):
Well done, I think its great that we keep an eye on Sonarqube and take action when errors or potential code smells arise!

PR #1282, opened 2024-03-08
INTYGFV-16405 & INTYGFV-16424: Only fetch ICF when needed

SamuelTilly:
__Fixes__

* Component re-rendering caused by imprecise `useSelector`
  - Certificate
  - CertificatePage
  - Question
  - QuestionWithSubQuestion
  - CertificateSidePanel
* Unnecessary state updated due to listening for `updateCertificate` redux action.

-------
To make sure that components are not re-render unnecessary, it helps to use `isEqual` or more precise selectors - for example:
```
const certificateId = useSelector((state) => state.ui.uiCertificate.certificate?.metadata.id)
```

instead of
```
const metadata = useSelector((state) => state.ui.uiCertificate.certificate?.metadata)
const certificateId = metadata.id
```

These updates can be identified using react DevTools extension for chrome.

`updateCertificate` action was previously only called when a certificate was initially loaded. This has changed to be used when validating certificates after updating fields. Middleware for SRS and FMB used for the side panel has been listening for this event and treating it as the initial load event for certificates.

### DIFF
--- a/apps/webcert/package.json
+++ b/apps/webcert/package.json
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@frontend/filtrex": "workspace:*",
+    "@frontend/utils": "workspace:^",
     "@reduxjs/toolkit": "1.7.2",
     "axios": "^1.6.0",
     "classnames": "^2.3.2",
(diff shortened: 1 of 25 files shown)


mhornfeldt (approved):
Nice. If other parts than what is mentioned in the jira was updated it can be added to the description of the jira so test can decide what tests they want to perform.

PR #1288, opened 2024-03-25
INTYGFV-16459: Added logic that redirects nurses to the same page after successful login as doctors.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks good! Nice that it's the default behaviour as long as the user aint a care admin.

PR #1289, opened 2024-03-26
INTYGFV-16466 & INTYGFV-16400: Add save button and hide print functionality for app

SamuelTilly:
- Added button and warning for saving PDF directly (INTYGFV-16466)
- Disable print functionality when using Inera App (INTYGFV-16400)
- Added information text about disabled print functionality
- Added `LinkButton` component that uses IDS CSS to create a link that is styled as a button.
- Dialog now gets aria-label by its headline.
- Added  `useAvailableFunction` hook for getting specific `availableFunction` from the store directly.


mhornfeldt (approved):
Nice!

PR #1293, opened 2024-04-08
INTYGFV-16486: Added signed attribute to CertificateMetadata and updated UvCareUnitAddress to handle signed if present.

ChrilleSigerhed:


ChrilleSigerhed:
image

PR #1298, opened 2024-04-15
INTYGFV-16530: Loop on failing user request

SamuelTilly:


ChrilleSigerhed:
Can we connect this to a jira so that test are able to verify the change?

SamuelTilly:
> Can we connect this to a jira so that test are able to verify the change?

Valid point, added a ticket

ChrilleSigerhed (approved):
Nice!

PR #1302, opened 2024-04-16
INTYGFV-16524: UeCheckboxDateRangeList should display label if present.

ChrilleSigerhed:
image


mhornfeldt (approved):
Approved! Reflected that it is a bit odd that the component didn't display label before since it's not introduced now for the component.

mhornfeldt:
We can check if label can be added in Question component instead

PR #1306, opened 2024-04-24
INTYGFV-16519: Added limitations min and max for date range list 

mhornfeldt:


apps/webcert/src/utils/certificate/getClientValidationErrors.ts
```diff
@@ -123,12 +132,51 @@ const getErrorsFromConfig = (id: string, config: CertificateDataConfigType, valu
...
+          .filter((val) => isDateBeforeLimit(config.min, val.to))
+          .map((val) =>
+            getValidationErrorWithCustomField(
+              id,
+              `Ange ett datum som är tidigast ${config.min ?? ''}.`,
+              'INVALID_FORMAT',
+              true,
+              `${val.id}.to`
+            )
+          )
```
SamuelTilly, line 177: This duplication here could perhaps be removed with a less generic function than `getValidationErrorWithCustomField`?
 
We could name the error type more specific, like "DATE_TO_EARLY" or something similar, since the date format is not really invalid here.

There is a `max` added to the config, should that also be validated?

mhornfeldt: I've asked förvaltningen if they want the same for max

apps/webcert/src/utils/certificate/getClientValidationErrors.test.ts
```diff
@@ -186,6 +186,37 @@ describe('Validation based on value', () => {
 })
 
 describe('Validation based on config', () => {
+  describe(`${ConfigTypes.UE_CHECKBOX_DATE_RANGE_LIST}`, () => {
```
SamuelTilly, line 189: Hmm, a bit confused about this test. `fakeDateElement` does not return a `UE_CHECKBOX_DATE_RANGE_LIST` it returns a `UE_DATE`. Not sure how `config.minDate` gives this result when `isDateBeforeLimit` is looking a `config.min`.
😕

mhornfeldt: My bad copy paste error

apps/webcert/src/utils/certificate/getClientValidationErrors.ts
mhornfeldt: Good comments I fixed everything but the max since I'm waiting for a response.

SamuelTilly: I thought about the `getValidationErrorWithCustomField` a bit more and I think we could use the existing `getValidationErrorFactory` function instead.

We could add something like:
```typescript
const getEarlyDateError = (id: string, field: string, min?: string) => {
  return getValidationErrorFactory(
    id,
    field
  )({ type: 'INVALID_FORMAT', text: `Ange ett datum som är tidigast ${min ?? ''}.`, showAlways: true })
}
```

And use it like this:
```typescript
const fromViolationAgainstMinLimit = value.list
          .filter((val) => isDateBeforeLimit(config.min, val.from))
          .map((val) => getEarlyDateError(id, `${val.id}.from`, config.min))
```

And UE_DATE could be simplified a bit more:
```typescript
case ConfigTypes.UE_DATE:
      if (value.type === CertificateDataValueType.DATE) {
        return isDateBeforeLimit(config.minDate, value.date) ? [getEarlyDateError(id, id, config.minDate)] : []
      }
      break
```

mhornfeldt: Tried this but the field doesn't get set. Will keep looking into it

mhornfeldt: Done and added max validation since I didn't get a response

SamuelTilly (approved):
Very nice 💯

PR #1312, opened 2024-05-06
INTYGFV-16525: Updated build-image and runtime-image versions

mikeyOne:


mww-sopra-steria (approved):
Nice! You can bump the middle number of the runtime image if you prefer.

PR #1322, opened 2024-05-21
INTYGFV-15929: Updated build-image and runtime-image and set servert_…

mikeyOne:
…tokens to off for maintenance.


ChrilleSigerhed (approved):
Nice!

PR #1323, opened 2024-05-31
INTYGFV-16646: Update cannot complement modal

SamuelTilly:


mww-sopra-steria (approved):
Nice! Works like a charm!

PR #1324, opened 2024-05-31
Typescript error in getResourceLink function

SamuelTilly:
```typescript
export const getResourceLink = (resourceLinks: ResourceLink[], wanted: ResourceLinkType): ResourceLink =>
  resourceLinks.find((link) => link.type === wanted)!
```

This function is ignoring cases where it could return `undefined` and is widely used in the application. This breaks predictability with typescript.


mhornfeldt (approved):
Nice :)

PR #1327, opened 2024-06-11
DEVENV: Added general maintenance page

mhornfeldt:


SamuelTilly:
Looks really good, well done!

A couple of things for the general site.

1. Shouldn't the top icon be red?
2. The footer should be at the bottom of the page.

image

mhornfeldt:
> Looks really good, well done!
> 
> A couple of things for the general site.
> 
> 1. Shouldn't the top icon be red?
> 2. The footer should be at the bottom of the page.
> 
> image

In the sketch it shows a red triangle which is named "attention" but this is the icon I get when I use "attention". This is just a temporary page since we're short on time so maybe it's okay with this icon this time.

SamuelTilly (approved):

PR #1332, opened 2024-06-13
INTYGFV-16672: get messageTypes from certificate metadata

SamuelTilly:


apps/webcert/src/types/certificate.ts
```diff
@@ -41,6 +42,7 @@ export interface CertificateMetadata {
   latestMajorVersion: boolean
   responsibleHospName: string
   signed?: string
+  messageTypes?: QuestionType[]
```
ChrilleSigerhed, line 45: MessageType is not an array of enums. The new type has the following structure:

```
[
 {
  type: QuestionType
  subject: string
 },
{
  type: QuestionType
  subject: string
 },
]
```

ChrilleSigerhed (changes requested):
Added comment regarding newly introduced type MessageType.

PR #1349, opened 2024-07-02
INTYGFV-16503: Improve validationError handling for UeDiagnoses

SamuelTilly:


ChrilleSigerhed (approved):
Works like a charm, well done.

Please wait with merging until we have merged the common/core parts so that we dont get a weird behaviour in devtest.

PR #1357, opened 2024-07-08
INTYGFV-15819: Navigate back button issues

SamuelTilly:
* remove unrecommended "connected-react-router"
* use `goBack` when trying to replace `pathname` with same as previous (avoid having to press back twice).
* move history from middlewares to a hook
* simplify `NavigateBackButton`

This should also simplify upgrading to react-router 6 and react 18 in the future.


mikeyOne (approved):
Looks reasonable to me :)

PR #1363, opened 2024-07-12
INTYGFV-16650: ResponsibleHospName should only be displayed if user origin is deep integration

mikaelaromild1:


apps/webcert/src/faker/user/fakeUser.ts
```diff
@@ -16,5 +16,6 @@ export const fakeUser = (value?: Partial<User>): User => ({
   loginMethod: LoginMethod.FAKE,
   protectedPerson: false,
   careProviders: [fakeCareProvider({ id: unit.unitId, name: unit.unitName })],
+  origin: origin,
```
mhornfeldt, line 19: Communicate condition through resource link as discussed so the origin does not reach the frontend

PR #1371, opened 2024-08-30
INTYGFV-16800: Added paths /v2 /webcert and /services to nginx config

mikeyOne:


mww-sopra-steria (changes requested):
Great that you added /v2 and /webcert for all the launch-apis!

Regarding the /services, I don't think we should expose them through the frontend reverse proxy. Shouldn't /services be accessed through the "ntjp"-route and with SITHS-certificates?

mww-sopra-steria (approved):

PR #1378, opened 2024-09-04
K1J-148: Trigger logout and cancelLogout only when djupintegration

mikeyOne:


mww-sopra-steria (approved):
Looks good to me! Should do the trick!

apps/webcert/src/App.tsx
```diff
@@ -10,23 +10,27 @@ import SubscriptionWarningModal from './feature/subscription/SubscriptionWarning
...
 import { ErrorCode, ErrorType } from './store/error/errorReducer'
-import { useAppDispatch } from './store/store'
+import { useAppDispatch, useAppSelector } from './store/store'
 import { cancelLogout, triggerLogout } from './store/user/userActions'
 import { initateApplication } from './store/welcome/welcomeActions'
+import { getUser } from './store/user/userSelectors'
 
 function App(): JSX.Element {
   const dispatch = useAppDispatch()
+  const origin = useAppSelector(getUser)?.origin
```
SamuelTilly, line 20: This could cause unnecessary renders, please change to `useAppSelector((state) => getUser(state).origin)`

mikeyOne: Thanks, I will make the change

PR #1380, opened 2024-09-12
K1J-156: Adjusted redirect via go back button after signing certifica…

mikeyOne:
…te with siths in fristaende wc


apps/webcert/src/hooks/useGoBackEffect.ts
```diff
@@ -0,0 +1,22 @@
+import { useHistory } from 'react-router'
+import { useAppSelector } from '../store/store'
+import { getPatientId } from '../store/srs/srsSelectors'
+
+export const useGoBackEffect = () => {
+  const history = useHistory()
+  const patientId = useAppSelector((state) => getPatientId(state))
+  const isAfterSithsSign = document.referrer.startsWith('https://esign')
```
SamuelTilly, line 8: Instead of doing document.location.reload() we could perhaps put this in a state that we "consume"

```
  const [isForreignReferrer, updateForreignReferrer] = useState(document.referrer.startsWith('https://esign'))

  return () => {
    if (isForreignReferrer) {
      if (patientId) {
        history.push(`/create/${encodeURIComponent(btoa(patientId))}`)
      } else {
        history.push('/search')
      }
      updateForreignReferrer(false)
    } else {
      history.goBack()
    }
  }
```

If this doesn't work, I would prefer if we use `window.location.replace` instead of `history.push` and remove `document.location.reload`.

SamuelTilly (changes requested):

apps/webcert/src/hooks/useGoBackEffect.ts
SamuelTilly: If we also check referrer against `window.location.hostname` we could perhaps make this a bit more general, supporting whatever domain may be in `document.referrer`

apps/webcert/src/hooks/useGoBackEffect.ts
```diff
@@ -0,0 +1,22 @@
+import { useHistory } from 'react-router'
+import { useAppSelector } from '../store/store'
+import { getPatientId } from '../store/srs/srsSelectors'
+
+export const useGoBackEffect = () => {
```
SamuelTilly, line 5: "Effect" is reserved for hooks that utilize "useEffect" to sync with external systems. Since this hook is not using that, we can remove that part from the name.

SamuelTilly (approved):

PR #1381, opened 2024-09-12
K1J-203: Simplify Tabs component and avoid updating undefined elements

SamuelTilly:


mww-sopra-steria (approved):
Nice! Hope it solves the issue!

PR #1383, opened 2024-09-13
K1J-204: Make sure to reset certificate state when changing unit

SamuelTilly:


ChrilleSigerhed (approved):
Impressive, well done!

PR #1384, opened 2024-09-13
K1J-62: Change to get availableToPatient status from metadata

mikeyOne:


mww-sopra-steria (approved):
Looks good! Nice to get read of some hardcoding in frontend!

PR #1398, opened 2024-09-24
K1J-104: Added users for SIT1

mhornfeldt:


mww-sopra-steria (approved):
I was thinking that we should do this through the testability-api, but when I see this solution it looks pretty pragmatic. Lets go with this for now!

PR #1400, opened 2024-09-25
K1J-289: ExpandableText link should not reload application

SamuelTilly:


apps/webcert/src/components/utils/ExpandableText.tsx
```diff
@@ -1,52 +1,42 @@
...
-
-export const ExpandableText: React.FC<Props> = ({ text, maxLength, additionalStyles }) => {
-  const [expand, setExpand] = useState(false)
-
-  useEffect(() => {
-    setExpand(false)
-  }, [text])
+}) {
+  const isExpandable = text && text.length > maxLength
+  const [expand, setExpand] = useState(isExpandable ? false : true)
```
ChrilleSigerhed, line 22: This could be simplified

```const [expand, setExpand] = useState(!isExpandable)```

If u feel that this makes it less readable, ignore this comment.

ChrilleSigerhed (approved):
Nice

apps/webcert/src/components/utils/ExpandableText.tsx
SamuelTilly: Good call, i'll fix this

PR #1401, opened 2024-09-25
INTYGFV-14695: Updated nginx configuration to allow /v1

ChrilleSigerhed:


apps/webcert/nginx/templates/default.conf.template
```diff
@@ -71,6 +71,16 @@ server {
       proxy_set_header X-Forwarded-Port $server_port;
     }
 
+    location /v1 {
```
SamuelTilly, line 74: Would a regex work for supporting multiple versions
`location ~ ^/v[12]` for 1 and 2 or `location ~ ^/v[0-9]+` for any version.

ChrilleSigerhed: I think its better that we allow certain endpoints explicity?

SamuelTilly: The first option `location ~ ^/v[12]` would be explicit. :)

PR #1408, opened 2024-09-30
Handle seed option with fakerFromSchema

SamuelTilly:
This makes it so that all subsequent call to the return function of `fakerFromSchema` with the seed option is predictable but doesn't produce the same result.

Before:
```typescript
const fakePerson = fakerFromSchema(z.object({
  firstName: z.string(),
}), { seed: 1 })

fakePerson() // Mina
fakePerson() // Mina
```

After:
```typescript
const fakePerson = fakerFromSchema(z.object({
  firstName: z.string(),
}), { seed: 1 })

fakePerson() // Mina
fakePerson() // Caesar
```


mhornfeldt (approved):
nice!

PR #1419, opened 2024-10-23
K1J-443: Fixed color for SRS risk lvl 2

mhornfeldt:


ChrilleSigerhed (approved):
Nice

apps/rehabstod/tailwind.config.ts
```diff
@@ -17,6 +17,7 @@ export default {
     colors: {
       black: '#000',
       white: '#FFF',
+      orange: '#D17200',
```
SamuelTilly, line 20: Looking at figma it looks like this color should be `#FF9517` same as "IDS-COLOR-GRAPHIC"

PR #1420, opened 2024-10-23
K1J-442: Use testability api for fake-login in Rehabstöd

SamuelTilly:
* Update fake login and logout to use new endpoints
* Update endpoints for fetching HSA data
* Refactoring welcome components
* Added more tests
* Lazy load welcome when needed (should not be part of the general bundle)


mww-sopra-steria (approved):
Looks and works great!

PR #1421, opened 2024-10-25
K1J-442: Extra rehabstod welcome fixes

SamuelTilly:


mww-sopra-steria (approved):
Nice! 

Could you add the following to default.conf.template for rehabstod?
/saml2
/login

And remove the following?
/fake
/saml

PR #1439, opened 2024-12-11
K1J-559: Fix issue with numbers beeing displayed with wrong state. On…

ChrilleSigerhed:
…ly display contactLenses if present on print.


mww-sopra-steria (approved):
Looks good!

PR #1445, opened 2024-12-18
K1J-632: Added banner for inactive certificate types.

ChrilleSigerhed:


ChrilleSigerhed:
image

apps/webcert/src/page/CertificatePage.tsx
```diff
@@ -82,6 +83,7 @@ const CertificatePage: React.FC = () => {
         !isCertificateDeleted && (
           <>
             <MajorVersionNotification />
+            <InactiveCertificateTypeNotification />
```
mww-sopra-steria, line 86: The comment is not on the correct line, but I add it here as MajorVersionNotification is added here.

Could we add logic to MajorVersionNotification that it is only displayed if inactiveCertificateType not true?

mww-sopra-steria (changes requested):
See comment. It will remove the risk of it displaying multiple notifications. The certificate could be of a version that is not the latest major of the type and it could happen that the type has been inactivated.

mww-sopra-steria (approved):
Nice!

PR #1450, opened 2025-02-12
K1J-776: decorate certificate data twice to assert we validate at correct depth

ChrilleSigerhed:


apps/webcert/src/store/certificate/certificateMiddleware.ts
```diff
@@ -894,11 +894,15 @@ const handleUpdateCertificateDataElement: Middleware<Dispatch> =
...
-            },
+            getDecoratedCertificateData(
+              {
+                ...certificate.data,
+                [action.payload.id]:
+                  validationErrors.length > 0 ? { ...element, validationErrors } : { ...action.payload, validationErrors: [] },
+              },
+              certificate.metadata,
+              certificate.links
+            ),
```
SamuelTilly, line 905: I would like to see a test for this. preferably for the getDecoratedCertificateData function it self. I tried adding one from the description of the problem but was unable to have it fail, perhaps im missing something.

```typescript
it('Should set visible to true for chain of dependending elements', () => {
  const { data, metadata, links } = fakeCertificate({
    data: {
      ...fakeRadioBooleanElement({
        id: '1',
        value: { id: 'first', selected: true },
      }),
      ...fakeRadioBooleanElement({
        id: '2',
        value: { id: 'second', selected: true },
        visible: false,
        validation: [
          fakeShowValidation({
            questionId: '1',
            expression: 'first',
          }),
        ],
      }),
      ...fakeRadioBooleanElement({
        id: '3',
        value: { id: 'third', selected: true },
        visible: false,
        validation: [
          fakeShowValidation({
            questionId: '2',
            expression: 'second',
          }),
        ],
      }),
    },
  })

  expect(getDecoratedCertificateData(data, metadata, links)['1'].visible).toBe(true)
  expect(getDecoratedCertificateData(data, metadata, links)['2'].visible).toBe(true)
  expect(getDecoratedCertificateData(data, metadata, links)['3'].visible).toBe(true)
})
```

SamuelTilly (changes requested):
Missing test

apps/webcert/src/store/certificate/certificateMiddleware.ts
SamuelTilly: Adding one more element looks like it triggers the problem

SamuelTilly: If we update `validateData` to evaluate that no more changes can occure to the state we can make this work indendently of depth.

```typescript
function validateData(data: CertificateData, metadata: CertificateMetadata): CertificateData {
  let previousData: CertificateData
  let currentData = data

  if (shouldBeReadOnly(metadata)) {
    return data
  }

  do {
    previousData = currentData
    currentData = Object.fromEntries(Object.entries(previousData).map(([id, element]) => [id, validateElement(previousData, element)]))
  } while (!isEqual(previousData, currentData))

  return currentData
}
```

mww-sopra-steria (approved):

PR #1460, opened 2025-04-14
K1J-586: Update secondary button text if user can not sign.

ChrilleSigerhed:


apps/webcert/src/feature/certificate/Buttons/SignAndSendButton.tsx
```diff
@@ -54,6 +54,7 @@ const SignAndSendButton: React.FC<Props> = ({
         open={confirmModalOpen}
         hideConfirmButton={!canSign}
         setOpen={setConfirmModalOpen}
+        declineButtonText={canSign ? 'Avbryt' : 'Stäng'}
```
mhornfeldt, line 57: This solution will potentially affect other modals than SIGN_CERTIFICATE_CONFIRMATION, then I'm not sure it's isolated to DB/DOI as we told förv. Or maybe we can be sure db/doi is the only one missing the sign link?

ChrilleSigerhed:
image

image

apps/webcert/src/feature/certificate/Buttons/SignAndSendButton.tsx
mhornfeldt: SignConfirmationModal is used by DB age check, but it has two buttons so it won't be affected this time.

mhornfeldt (approved):

PR #1462, opened 2025-04-15
K1J-969: Refactor UvDateRange component and corresponding tests

ChrilleSigerhed:
image

### DIFF
--- a/apps/webcert/src/feature/certificate/UnifiedEdit/UeDateRange/UeDateRange.tsx
+++ b/apps/webcert/src/feature/certificate/UnifiedEdit/UeDateRange/UeDateRange.tsx
@@ -7,16 +7,16 @@ import { updateCertificateDataElement } from '../../../../store/certificate/cert
 import { getVisibleValidationErrors } from '../../../../store/certificate/certificateSelectors'
 import DatePickerCustom from '../../../../components/Inputs/DatePickerCustom/DatePickerCustom'
 import QuestionValidationTexts from '../../../../components/Validation/QuestionValidationTexts'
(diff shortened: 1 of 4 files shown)


mhornfeldt:
We need to update the field ids so it follows CS like for date range list

apps/webcert/src/feature/certificate/UnifiedEdit/UeDateRange/UeDateRange.tsx
```diff
@@ -165,7 +165,9 @@ const UeDateRange: React.FC<Props> = ({ question, disabled }) => {
             textInputOnBlur={handleToTextInputOnBlur}
             textInputOnKeyDown={handleToTextInputOnKeyDown}
             textInputDataTestId={`tom${config.id}`}
-            displayValidationErrorOutline={getShouldDisplayValidationErrorOutline(config.id, 'tom')}
+            displayValidationErrorOutline={
+              getShouldDisplayValidationErrorOutline(config.id, 'tom') || getShouldDisplayValidationErrorOutline(config.id, 'to')
+            }
```
mhornfeldt, line 170: I think there was a marking for the whole period that was missing too, besides "to".

mhornfeldt (approved):

apps/webcert/src/feature/certificate/UnifiedEdit/UeDateRange/UeDateRange.tsx
ChrilleSigerhed: Nice catch! Updated and now the error is displayed correclty

image

PR #1464, opened 2025-04-16
K1J-998: Introduce JSON-logging based on ECS in frontend apps. Generate trace-id from frontend for traceability of calls from frontend to backend services. Change create draft api call to send sensitive data in payload instead of query parameters. 

mww-sopra-steria:


ChrilleSigerhed (approved):
Well done!

PR #1470, opened 2025-04-22
K1J-270: logout idp as session ping effect

SamuelTilly:


ChrilleSigerhed (approved):
Well done, works like a charm!

PR #1493, opened 2025-08-12
K1J-1225: fix issue with loading previous certificates

SamuelTilly:


ChrilleSigerhed (commented):
Well done, seems like this fix has solved the issue with patient id not beeing correctly sent in the filter!

However, when I check the dev tools window, it seems like we are making mulitple similar requests when renderingen the previous certificate page:

Image

It that something we should follow up and try to solve in this PR aswell, or how do you think its best handled?

PR #1495, opened 2025-08-12
K1J-1240: Add icfCodesPropertyName

alryabov:


apps/webcert/src/store/icf/icfSelectors.ts
```diff
@@ -1,10 +1,11 @@
 import type { RootState } from '../store'
 import type { AvailableIcfCodes } from './icfReducer'
+import { IcfCodesPropertyEnum } from '../../types'
 
 export const getIcfData =
-  (id: string) =>
+  (icfCodesPropertyName: IcfCodesPropertyEnum) =>
   (state: RootState): AvailableIcfCodes | undefined => {
-    return id === 'aktivitetsbegransning' ? state.ui.uiIcf.activityLimitation : state.ui.uiIcf.disability
+    return state.ui.uiIcf[IcfCodesPropertyEnum[icfCodesPropertyName]];
```
SamuelTilly, line 8: This results in a typescript error since typescript is unable to infer what `IcfCodesPropertyEnum[icfCodesPropertyName]` resolves to. An alternative could be to use an object like so:

```typescript
const IcfCodesProperty = {
  FUNKTIONSNEDSETTNINGAR: 'disability',
  AKTIVITETSBEGRANSNINGAR: 'activityLimitation',
} as const

export const getIcfData =
  (icfCodesPropertyName: keyof typeof IcfCodesProperty) =>
  (state: RootState): AvailableIcfCodes | undefined => {
    return state.ui.uiIcf[IcfCodesProperty[icfCodesPropertyName]]
  }
```

SamuelTilly (changes requested):

PR #1500, opened 2025-08-20
K1J-984: Fix minor IDS issues for rehabstod

SamuelTilly:


mhornfeldt (approved):
Looks good :) Maybe some tests should be added if it's not already full coverage

PR #1618, opened 2026-04-15
K1J-1522-3: seperate wc from general components

LucasLundstrom:
testing build on server..


ChrilleSigerhed (approved):
Awsome work, good that u kept the "common components" isolated in the package commonComponents so that we can later decide how we want to restructure these in our frontend mono repo.

PR #1624, opened 2026-05-12
K1J-1986: Add filter to unsupported characters and display info box informing user of sanitazion.

ChrilleSigerhed:


alryabov (approved):
Looks great!

PR #1634, opened 2026-05-29
K1J-2145: Validate case communication messages

jacobalmqvistsopra:


apps/webcert/src/components/question/QuestionForm.tsx
```diff
@@ -43,10 +45,11 @@ const QuestionForm = ({ questionDraft }: Props) => {
   const isMissingType = useAppSelector(isQuestionMissingType)
   const isMissingMessage = useAppSelector(isQuestionMissingMessage)
   const showValidationMessages = useAppSelector(isDisplayValidationMessages)
-  const [message, setMessage] = useState(questionDraft.message)
   const messageTypes = useAppSelector(getCertificateMessageTypes, shallowEqual)
   const isFunctionDisabled = useAppSelector(isQuestionFunctionDisabled)
   const MAX_NUMBER_OF_ALLOWED_CHARACTERS: number = 4999
+  const { sanitize, showWarning, sanitizedInitialValue: sanitizedInitialDesc } = useIso8859Sanitization(questionDraft.message)
```
ChrilleSigerhed: We need to sanitize the message for drafts from backend aswell, you can check ```UeTextField``` for inspiration. Otherwise if the user loads a question with a message containing special characters already persisted we get the following outcome that will result in a technical error when they try to send the message:

Image

ChrilleSigerhed (changes requested):
Nice!

Added some comments regarding cleanup from already persisted questions from backend. Also, it would be very nice if you added some tests to verify this logic.

apps/webcert/src/components/question/QuestionForm.tsx
jacobalmqvistsopra: Good catch, will fix.

ChrilleSigerhed (approved):

PR #1635, opened 2026-06-01
K1J-2145: Clear info box when message is sent

jacobalmqvistsopra:


ChrilleSigerhed (approved):
Looks good, I assume the tests for this will also come in the "test-pr" we discuessed yesterday

